PR #84, opened 2026-03-03
Add installable CMake package smoke test

petlenz:
## Summary

  This PR tightens the install/package path for `NumSim_CAS` and adds downstream validation for it.

  ## Changes

  - install generated `NumSim_CASConfig.cmake`, version file, and exported targets
  - bundle the required `tmech` headers and a minimal fallback `tmech` package config into the install prefix
  - add a `numsim_cas_package_smoke` CTest that installs the project into a temporary prefix, configures a separate consumer with `find_package(NumSim_CAS CONFIG REQUIRED)`, builds it,
  and runs it

  ## Why

  - catches packaging regressions that normal unit tests would miss
  - validates installed-consumer usage, not just in-tree builds
  - improves CI confidence for release/install workflows

  ## Validation

  - `ctest --test-dir build --output-on-failure`
  - full suite passes locally, including the new package smoke test


petlenz:
Skipping in this batch: 85 commits behind main, 4 CI failures on macOS Debug + Windows Debug. The failures are platform-specific (likely install paths or RPATH) and shouldn't be admin-merged without investigation. Needs:
1. Rebase onto main (significant — 85-commit delta).
2. Re-run CI and diagnose the platform failures.
3. Fix or document as known limitation.

Will revisit after the rest of the backlog is cleared.

CMakeLists.txt
```diff
@@ -167,7 +172,12 @@ endif()
 # -----------------------------
 # Install
 # -----------------------------
-if(NUMSIM_CAS_INSTALL_LIBRARY)
+set(NUMSIM_CAS_ENABLE_PACKAGE_SMOKE_TEST OFF)
+if(NUMSIM_CAS_INSTALL_LIBRARY AND NUMSIM_CAS_TMECH_IS_EXTERNAL_PACKAGE)
```
petlenz, line 273: **Breaking change.** The original `if(NUMSIM_CAS_INSTALL_LIBRARY)` block ran regardless of whether tmech was found externally. Conjoining with `NUMSIM_CAS_TMECH_IS_EXTERNAL_PACKAGE` means a user who sets `NUMSIM_CAS_INSTALL_LIBRARY=ON` while using the default FetchContent'd tmech now gets no install at all — silently, with only a STATUS message at the `elseif` below.

Recommend either (a) keeping `install(TARGETS ...)` outside this gate and only conditionalising the config-file generation, or (b) upgrading the `elseif` message to `WARNING` so the regression is visible.

CMakeLists.txt
```diff
@@ -186,12 +196,33 @@ if(NUMSIM_CAS_INSTALL_LIBRARY)
...
+
+  install(
+    FILES
+      "${CMAKE_CURRENT_BINARY_DIR}/NumSim_CASConfig.cmake"
+      "${CMAKE_CURRENT_BINARY_DIR}/NumSim_CASConfigVersion.cmake"
+    DESTINATION "${NUMSIM_CAS_INSTALL_CMAKEDIR}"
+  )
+elseif(NUMSIM_CAS_INSTALL_LIBRARY)
+  message(STATUS
+    "NUMSIM_CAS_INSTALL_LIBRARY is ON, but installable package export is disabled "
```
petlenz: `STATUS` messages are easily missed in CMake's normal output (especially via IDE configure steps). Use `WARNING` here — a user who set `NUMSIM_CAS_INSTALL_LIBRARY=ON` expected an install and now isn't getting one.

tests/package_smoke/package_smoke_test.cmake
```diff
@@ -0,0 +1,63 @@
...
+
+file(REMOVE_RECURSE "${_install_prefix}" "${_consumer_build_dir}")
+file(MAKE_DIRECTORY "${_install_prefix}" "${_consumer_build_dir}")
+
+execute_process(
+  COMMAND "${CMAKE_COMMAND}" --install "${NUMSIM_CAS_BINARY_DIR}" --prefix "${_install_prefix}" ${_config_args}
+  RESULT_VARIABLE _install_result
+)
+if(NOT _install_result EQUAL 0)
+  message(FATAL_ERROR "Package smoke test failed during install step")
```
petlenz: All three `execute_process` calls in this script (install, configure, build) fail with no captured output. When this test breaks on Windows/macOS CI, the next person to debug it will have only `FATAL_ERROR: Package smoke test failed during install step` to go on.

Capture both streams and dump on failure:
```cmake
execute_process(
  COMMAND "${CMAKE_COMMAND}" --install "${NUMSIM_CAS_BINARY_DIR}" --prefix "${_install_prefix}" ${_config_args}
  RESULT_VARIABLE _install_result
  OUTPUT_VARIABLE _install_stdout
  ERROR_VARIABLE _install_stderr
)
if(NOT _install_result EQUAL 0)
  message(FATAL_ERROR "Package smoke test install step failed:\n${_install_stdout}\n${_install_stderr}")
endif()
```

tests/package_smoke/package_smoke_test.cmake
```diff
@@ -0,0 +1,63 @@
...
+
+set(_consumer_exe "${_consumer_build_dir}/numsim_cas_package_smoke")
+if(DEFINED TEST_CONFIG AND NOT TEST_CONFIG STREQUAL "")
+  if(EXISTS "${_consumer_build_dir}/${TEST_CONFIG}/numsim_cas_package_smoke"
+     OR EXISTS "${_consumer_build_dir}/${TEST_CONFIG}/numsim_cas_package_smoke.exe")
+    set(_consumer_exe "${_consumer_build_dir}/${TEST_CONFIG}/numsim_cas_package_smoke")
+  endif()
+endif()
+if(CMAKE_HOST_WIN32)
+  string(APPEND _consumer_exe ".exe")
```
petlenz, line 72: The `.exe` is appended unconditionally for Windows hosts AFTER the multi-config branch has already set `_consumer_exe` to a path that may or may not have `.exe`. Worth a trace through:

- Windows + single-config: `_consumer_exe` = `<build>/numsim_cas_package_smoke`, then `.exe` appended → `<build>/numsim_cas_package_smoke.exe`. ✅
- Windows + multi-config: the `EXISTS` check matches either `<build>/<config>/numsim_cas_package_smoke` (no ext) OR `<build>/<config>/numsim_cas_package_smoke.exe`, then unconditionally sets the no-ext form. The `.exe` append makes it correct. ✅
- Linux/macOS: `.exe` append doesn't fire. ✅

Logic is correct — but the round-trip through "set no-ext, then append .exe" is confusing. A clearer form:
```cmake
if(CMAKE_HOST_WIN32)
  set(_exe_suffix ".exe")
else()
  set(_exe_suffix "")
endif()
set(_consumer_exe "${_consumer_build_dir}/numsim_cas_package_smoke${_exe_suffix}")
if(DEFINED TEST_CONFIG AND NOT TEST_CONFIG STREQUAL "")
  set(_multi "${_consumer_build_dir}/${TEST_CONFIG}/numsim_cas_package_smoke${_exe_suffix}")
  if(EXISTS "${_multi}")
    set(_consumer_exe "${_multi}")
  endif()
endif()
```
Not blocking; just a readability win.

petlenz (commented):
## Critical review — installable CMake package smoke test

The overall shape is right: install config files, build a downstream consumer against the install tree, run it. CI is green. A few things to address before merge.

### Behavior change: `NUMSIM_CAS_INSTALL_LIBRARY` is now a no-op with FetchContent'd tmech

Before this PR, `NUMSIM_CAS_INSTALL_LIBRARY=ON` installed headers and the targets export, even with FetchContent'd tmech (just without the `NumSim_CASConfig.cmake` config file). After this PR, when tmech is not found as an external package, the entire `install(TARGETS ...)` block is skipped — only a STATUS message prints. Downstream users who rely on `cmake --install` to copy headers will find their workflow silently broken.

Two options:
1. Keep the prior behavior (always install headers/targets when `NUMSIM_CAS_INSTALL_LIBRARY=ON`; only gate the config-file generation on `NUMSIM_CAS_TMECH_IS_EXTERNAL_PACKAGE`).
2. Escalate the STATUS message to `WARNING` so users notice the silent skip.

Option 1 preserves the API surface; option 2 at least makes the regression visible. The current code does neither.

### Diagnostics: `execute_process` calls silently swallow stderr

Every `execute_process` in `tests/package_smoke/package_smoke_test.cmake` runs without `OUTPUT_VARIABLE` / `ERROR_VARIABLE`. On failure, all you get is `FATAL_ERROR: Package smoke test failed during install step` — no diagnostic. The next person to debug this on Windows or a different platform will have nothing to go on. Capture stdout/stderr and dump on non-zero exit.

### Test coverage gap: smoke runs only with external tmech

`NUMSIM_CAS_ENABLE_PACKAGE_SMOKE_TEST` requires `NUMSIM_CAS_TMECH_IS_EXTERNAL_PACKAGE`, so the default developer build (FetchContent'd tmech, per the recent #252 pin) doesn't exercise the install path at all. CI also doesn't appear to install tmech as a system package — verify whether ANY CI job currently runs the smoke test. If not, the lock-in is theoretical.

At minimum the README/CONTRIBUTING should mention that exercising this test requires a tmech system install.

### Nits

- The Windows `.exe` append at the bottom of `package_smoke_test.cmake` is correct in both single-config and multi-config cases. Could be simplified to set the right path inside the multi-config branch and skip the unconditional append, but the current logic is fine.
- The `_install_prefix`/`_consumer_build_dir` wipe-and-rebuild on every run is correct (avoids stale install artifacts) but makes the test slow. Acceptable trade-off.

PR #90, opened 2026-05-14
Remove flaky tensor diff seed skip list by fixing underlying bugs

### LINKED ISSUE #89: Fuzzy tensor diff seeds skipped due to underlying CAS bugs
## Background

`tests/FuzzyTensorDiffTest.h` previously carried a hard-coded skip list of seeds that crashed or produced bogus results during randomized tensor differentiation. The skip list silently hid several distinct bugs in the CAS core rather than fixing them.

## Bugs exposed by the failing seeds

1. **Null expression crashes in scalar differentiation.** `scalar_negative` and `scalar_pow` could call `is_same<scalar_zero>` on invalid `expr_holder_t` values, tripping the assert inside `is_same`.
2. **`is_same` asserted on invalid expressions** instead of returning `false`, propagating the crash.
3. **Duplicate child insertion in `n_ary_tree`.** `merge_add` (in `functions.h`) and the `n_ary_add_dispatch` path in `simplifier_add.h` could push a combined child (e.g. `x + x → 2*x`) that already existed in the result map, hitting `n_ary_tree::insert_hash`'s duplicate guard.
4. **Scalar values dropped in tensor-to-scalar evaluation.** `tensor_evaluator` for `tensor_to_scalar_with_tensor_mul` constructed a fresh `tensor_to_scalar_evaluator` but only forwarded tensor values, so any scalar symbol on the RHS was missing.
5. **`inv()` of a skew-symmetric tensor in odd dimensions.** Skew-symmetric matrices in odd dimensions are singular (det = 0), so taking the inverse is mathematically invalid. The CAS accepted these expressions and then produced NaNs at evaluation time.
6. **`trans(A) - A` (or `A - trans(A)`) not recognized as skew-symmetric** at construction time, so subsequent `inv()` over odd dim wasn't rejected.

## Goal

Fix the underlying bugs so the skip list can go away. Add a regression test for `inv(skew)` rejection.

## Affected files (high level)

- `include/numsim_cas/core/simplifier/simplifier_add.h`
- `include/numsim_cas/functions.h`
- `include/numsim_cas/basic_functions.h`
- `include/numsim_cas/scalar/visitors/scalar_differentiation.h`
- `src/numsim_cas/scalar/visitors/scalar_differentiation.cpp`
- `include/numsim_cas/tensor/tensor_operators.h`
- `include/numsim_cas/tensor/visitors/tensor_evaluator.h`
- `tests/FuzzyTensorDiffTest.h` (skip list removed)
- `tests/FuzzyDiffBase.h`

petlenz:
Closes #89.

## Summary

`FuzzyTensorDiffTest` carried a skip list of seeds that crashed or returned bogus results during randomized tensor differentiation. The skip list silently masked several distinct CAS-core bugs. This PR removes the skip list and fixes each underlying bug.

## What changed

**Core / scalar**
- `is_same<T>` now returns `false` for invalid expressions instead of asserting (`basic_functions.h`).
- `scalar_differentiation::apply_imp` resets `m_result` on entry and treats an unvisited result as zero (`scalar_differentiation.h`).
- `scalar_negative` / `scalar_pow` differentiation tolerate invalid `dg`/`dh` and dispatch the constant-exponent and constant-base branches explicitly (`scalar_differentiation.cpp`).

**n_ary add merging**
- `merge_add` (in `functions.h`) and `n_ary_add_dispatch` (in `simplifier_add.h`) now combine with any pre-existing entry when a freshly-combined child (e.g. `x + x → 2*x`) collides with the result map. Previously these hit the duplicate-child guard in `n_ary_tree`.

**Tensor / inv(skew)**
- `tensor_operators.h`: `trans(A) - A` and `A - trans(A)` are recognized as skew-symmetric at construction time (odd dim).
- `inv()` rejects expressions that contain a skew-symmetric factor in odd dimensions (singular by det = 0). Skew detection routes through the projector node rather than relying on space annotations, so it survives propagation through `tensor_mul`.

**Tensor-to-scalar evaluation**
- `tensor_evaluator` for `tensor_to_scalar_with_tensor_mul` now forwards scalar symbol values from the surrounding scalar evaluator into the inner `tensor_to_scalar_evaluator` (`tensor_evaluator.h`).

**Tests**
- Skip list removed from `FuzzyTensorDiffTest.h`.
- Regression coverage for `inv(skew)` rejection added to `FuzzyTensorDiffTest.h` / `FuzzyDiffBase.h`.

## Test plan

- [x] `cmake --build build` — clean build with no new warnings
- [x] `ctest` — 100% pass (1482/1482, 9 unrelated FuzzyScalar skips remain)
- [x] Previously-skipped FuzzyTensor seeds now exercised by the suite

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/include/numsim_cas/basic_functions.h
+++ b/include/numsim_cas/basic_functions.h
@@ -10,7 +10,8 @@ namespace numsim::cas {
 
 template <typename Type, typename Expr>
 [[nodiscard]] inline bool is_same(Expr const &expr) noexcept {
-  assert(expr.is_valid());
+  if (!expr.is_valid())
+    return false;
   return Type::get_id() == expr.get().id();
 }
 
--- a/include/numsim_cas/core/evaluator_base.h
+++ b/include/numsim_cas/core/evaluator_base.h
@@ -21,10 +21,6 @@ public:
     m_symbols_to_value[to_base_holder(symbol)] = val;
   }
 
-  [[nodiscard]] auto const &symbol_values() const noexcept {
-    return m_symbols_to_value;
-  }
-
 protected:
   void dispatch() {
     auto it = m_symbols_to_value.find(m_current_expr);
--- a/include/numsim_cas/core/evaluator_base.h
+++ b/include/numsim_cas/core/evaluator_base.h
@@ -21,6 +21,10 @@ public:
     m_symbols_to_value[to_base_holder(symbol)] = val;
   }
 
+  [[nodiscard]] auto const &symbol_values() const noexcept {
+    return m_symbols_to_value;
+  }
+
 protected:
   void dispatch() {
     auto it = m_symbols_to_value.find(m_current_expr);
--- a/include/numsim_cas/core/n_ary_tree.h
+++ b/include/numsim_cas/core/n_ary_tree.h
@@ -2,7 +2,6 @@
 #define N_ARY_TREE_H
 
 #include <algorithm>
-#include <iostream>
 #include <numsim_cas/core/cas_error.h>
 #include <numsim_cas/core/expression_holder.h>
 #include <numsim_cas/core/hash_functions.h>
@@ -138,20 +137,8 @@ private:
   expr_ordered_map<expr_holder_t> m_symbol_map;
 
 private:
-  void dump_duplicate_diag(expression_holder<expr_t> const &expr) const {
-    std::cerr << "[DIAG] n_ary_tree duplicate insertion\n"
-              << "  new expr hash=" << expr.get().hash_value()
-              << " id=" << expr.get().id() << "\n"
-              << "  existing map (" << m_symbol_map.size() << " entries):\n";
-    for (auto const &[k, v] : m_symbol_map) {
-      std::cerr << "    hash=" << v.get().hash_value() << " id=" << v.get().id()
-                << " eq=" << (v == expr) << "\n";
-    }
-  }
-
   void insert_hash(expression_holder<expr_t> const &expr) {
     if (m_symbol_map.contains(expr)) {
-      dump_duplicate_diag(expr);
       throw internal_error(
           "n_ary_tree::insert_hash: duplicate child insertion");
     }
@@ -160,7 +147,6 @@ private:
 
   void insert_hash(expression_holder<expr_t> &&expr) {
     if (m_symbol_map.contains(expr)) {
-      dump_duplicate_diag(expr);
       throw internal_error(
           "n_ary_tree::insert_hash: duplicate child insertion");
     }
--- a/include/numsim_cas/core/n_ary_tree.h
+++ b/include/numsim_cas/core/n_ary_tree.h
@@ -2,6 +2,7 @@
 #define N_ARY_TREE_H
 
 #include <algorithm>
+#include <iostream>
 #include <numsim_cas/core/cas_error.h>
 #include <numsim_cas/core/expression_holder.h>
 #include <numsim_cas/core/hash_functions.h>
@@ -137,8 +138,20 @@ private:
   expr_ordered_map<expr_holder_t> m_symbol_map;
 
 private:
+  void dump_duplicate_diag(expression_holder<expr_t> const &expr) const {
+    std::cerr << "[DIAG] n_ary_tree duplicate insertion\n"
+              << "  new expr hash=" << expr.get().hash_value()
+              << " id=" << expr.get().id() << "\n"
+              << "  existing map (" << m_symbol_map.size() << " entries):\n";
+    for (auto const &[k, v] : m_symbol_map) {
+      std::cerr << "    hash=" << v.get().hash_value() << " id=" << v.get().id()
+                << " eq=" << (v == expr) << "\n";
+    }
+  }
+
   void insert_hash(expression_holder<expr_t> const &expr) {
     if (m_symbol_map.contains(expr)) {
+      dump_duplicate_diag(expr);
       throw internal_error(
           "n_ary_tree::insert_hash: duplicate child insertion");
     }
@@ -147,6 +160,7 @@ private:
 
   void insert_hash(expression_holder<expr_t> &&expr) {
     if (m_symbol_map.contains(expr)) {
+      dump_duplicate_diag(expr);
       throw internal_error(
           "n_ary_tree::insert_hash: duplicate child insertion");
     }
--- a/include/numsim_cas/core/n_ary_tree.h
+++ b/include/numsim_cas/core/n_ary_tree.h
@@ -58,6 +58,12 @@ public:
   // different key (cos(x)^2 + sin(x)^2 -> 1, x + (-x) -> 0, projector
   // identities) which can collide with yet another entry in the map. Loop
   // until insertion produces no further collisions.
+  //
+  // Not atomic under exception: if `+` throws on iteration N>1, prior
+  // erasures are already committed. Current callers (merge_add,
+  // n_ary_add_dispatch) build a fresh result tree and discard it on throw,
+  // so the partial state is never observed. Do not call this from a context
+  // that catches the exception and continues using the tree.
   inline void merge_or_insert(expression_holder<expr_t> entry) {
     while (true) {
       auto it = m_symbol_map.find(entry);
--- a/include/numsim_cas/core/n_ary_tree.h
+++ b/include/numsim_cas/core/n_ary_tree.h
@@ -53,6 +53,23 @@ public:
     insert_hash(std::move(expr));
   }
 
+  // Insert `entry`, combining with any colliding map entry first.
+  // After combination, `+` may algebraically simplify to an expression with a
+  // different key (cos(x)^2 + sin(x)^2 -> 1, x + (-x) -> 0, projector
+  // identities) which can collide with yet another entry in the map. Loop
+  // until insertion produces no further collisions.
+  inline void merge_or_insert(expression_holder<expr_t> entry) {
+    while (true) {
+      auto it = m_symbol_map.find(entry);
+      if (it == m_symbol_map.end())
+        break;
+      auto combined = it->second + entry;
+      m_symbol_map.erase(it);
+      entry = std::move(combined);
+    }
+    insert_hash(std::move(entry));
+  }
+
   inline void reserve([[maybe_unused]] std::size_t size) noexcept {
     // m_symbol_map.reserve(size);
   }
--- a/include/numsim_cas/core/simplifier/simplifier_add.h
+++ b/include/numsim_cas/core/simplifier/simplifier_add.h
@@ -295,16 +295,7 @@ public:
     if (pos != add.symbol_map().end()) {
       auto expr{pos->second + base::m_rhs};
       add.symbol_map().erase(pos);
-      // Combined result may match another entry (e.g. x+x=2*x collides
-      // with existing 2*x when add has {x, 2*x})
-      auto pos2 = add.symbol_map().find(expr);
-      if (pos2 != add.symbol_map().end()) {
-        auto combined = pos2->second + expr;
-        add.symbol_map().erase(pos2);
-        add.push_back(std::move(combined));
-      } else {
-        add.push_back(std::move(expr));
-      }
+      add.merge_or_insert(std::move(expr));
       return expr_add;
     }
     add.push_back(base::m_rhs);
--- a/include/numsim_cas/core/simplifier/simplifier_add.h
+++ b/include/numsim_cas/core/simplifier/simplifier_add.h
@@ -295,8 +295,8 @@ public:
     if (pos != add.symbol_map().end()) {
       auto expr{pos->second + base::m_rhs};
       add.symbol_map().erase(pos);
-      // The combined result might match another entry in the add
-      // (e.g. add has {x, 2*x}, combining x+x=2*x collides with 2*x)
+      // Combined result may match another entry (e.g. x+x=2*x collides
+      // with existing 2*x when add has {x, 2*x})
       auto pos2 = add.symbol_map().find(expr);
       if (pos2 != add.symbol_map().end()) {
         auto combined = pos2->second + expr;
--- a/include/numsim_cas/core/simplifier/simplifier_add.h
+++ b/include/numsim_cas/core/simplifier/simplifier_add.h
@@ -295,7 +295,16 @@ public:
     if (pos != add.symbol_map().end()) {
       auto expr{pos->second + base::m_rhs};
       add.symbol_map().erase(pos);
-      add.push_back(std::move(expr));
+      // The combined result might match another entry in the add
+      // (e.g. add has {x, 2*x}, combining x+x=2*x collides with 2*x)
+      auto pos2 = add.symbol_map().find(expr);
+      if (pos2 != add.symbol_map().end()) {
+        auto combined = pos2->second + expr;
+        add.symbol_map().erase(pos2);
+        add.push_back(std::move(combined));
+      } else {
+        add.push_back(std::move(expr));
+      }
       return expr_add;
     }
     add.push_back(base::m_rhs);
--- a/include/numsim_cas/core/simplifier/simplifier_sub.h
+++ b/include/numsim_cas/core/simplifier/simplifier_sub.h
@@ -96,15 +96,7 @@ public:
   expr_holder_t dispatch(typename Traits::add_type const &rhs) {
     auto add_expr{make_expression<typename Traits::add_type>(rhs)};
     auto &add{add_expr.template get<typename Traits::add_type>()};
-    auto inner = lhs.expr();
-    auto pos = add.symbol_map().find(inner);
-    if (pos != add.symbol_map().end()) {
-      auto combined = pos->second + inner;
-      add.symbol_map().erase(pos);
-      add.push_back(combined);
-    } else {
-      add.push_back(inner);
-    }
+    add.merge_or_insert(lhs.expr());
     return make_expression<typename Traits::negative_type>(std::move(add_expr));
   }
 
@@ -216,6 +208,11 @@ public:
   }
 
   // merge two add expressions
+  // NOTE: this dispatch has a separate pre-existing semantic issue —
+  // `lhs.coeff() + rhs.coeff()` is unguarded against invalid coeffs and uses
+  // `+` where `-` would be correct for subtraction. Documented in
+  // CoreBugFixTest near SubSymbolDispatchSmoke. Conversion below preserves
+  // existing behavior and closes the transitive-collision bug class.
   expr_holder_t dispatch(typename Traits::add_type const &rhs) {
     auto expr{make_expression<typename Traits::add_type>()};
     auto &add{expr.template get<typename Traits::add_type>()};
@@ -225,15 +222,15 @@ public:
       auto pos{rhs.symbol_map().find(child)};
       if (pos != rhs.symbol_map().end()) {
         used_expr.insert(pos->second);
-        add.push_back(child + pos->second);
+        add.merge_or_insert(child + pos->second);
       } else {
-        add.push_back(child);
+        add.merge_or_insert(child);
       }
     }
     if (used_expr.size() != rhs.size()) {
       for (auto &child : rhs.symbol_map() | std::views::values) {
         if (!used_expr.count(child)) {
-          add.push_back(child);
+          add.merge_or_insert(child);
         }
       }
     }
@@ -250,7 +247,7 @@ public:
     if (pos != add.symbol_map().end()) {
       auto expr{pos->second - base::m_rhs};
       add.symbol_map().erase(pos);
-      add.push_back(expr);
+      add.merge_or_insert(std::move(expr));
       return expr_add;
     }
     add.push_back(-base::m_rhs);
--- a/include/numsim_cas/functions.h
+++ b/include/numsim_cas/functions.h
@@ -27,8 +27,10 @@ constexpr inline void merge_add(n_ary_tree<Derived> const &lhs,
     }
   }
 
-  // Helper: insert into result, combining with existing entry if present
-  auto insert_or_combine = [&](expression_holder<expr_t> entry) {
+  // Helper: insert into result, combining with existing entry if present.
+  // After combining child+rhs_child (e.g. x+x=2*x), the result may
+  // match another entry already in the result (e.g. existing 2*x).
+  auto safe_push = [&](expression_holder<expr_t> entry) {
     auto it = result.symbol_map().find(entry);
     if (it != result.symbol_map().end()) {
       auto combined = it->second + entry;
@@ -44,15 +46,15 @@ constexpr inline void merge_add(n_ary_tree<Derived> const &lhs,
     auto pos{rhs.symbol_map().find(child)};
     if (pos != rhs.symbol_map().end()) {
       used_expr.insert(pos->second);
-      insert_or_combine(child + pos->second);
+      safe_push(child + pos->second);
     } else {
-      insert_or_combine(child);
+      safe_push(child);
     }
   }
   if (used_expr.size() != rhs.size()) {
     for (auto &child : rhs.symbol_map() | std::views::values) {
       if (!used_expr.count(child)) {
-        insert_or_combine(child);
+        safe_push(child);
       }
     }
   }
--- a/include/numsim_cas/functions.h
+++ b/include/numsim_cas/functions.h
@@ -27,34 +27,20 @@ constexpr inline void merge_add(n_ary_tree<Derived> const &lhs,
     }
   }
 
-  // Helper: insert into result, combining with existing entry if present.
-  // After combining child+rhs_child (e.g. x+x=2*x), the result may
-  // match another entry already in the result (e.g. existing 2*x).
-  auto safe_push = [&](expression_holder<expr_t> entry) {
-    auto it = result.symbol_map().find(entry);
-    if (it != result.symbol_map().end()) {
-      auto combined = it->second + entry;
-      result.symbol_map().erase(it);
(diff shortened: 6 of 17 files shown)


petlenz:
Self-review of this PR — concerns worth addressing before merge.

## Scope mismatch with the description

The PR body says "`inv()` rejects expressions that contain a skew-symmetric factor in odd dimensions." That rejection lives entirely in `tests/FuzzyTensorDiffTest.h` (`is_provably_skew` / `contains_skew_factor`) as a fuzz-generator filter. The library still accepts and evaluates `inv(skew_3d)` and will produce NaN/Inf at runtime. Either:

- Move the check into `inv()` construction in `tensor_functions.h` (same pattern as the existing `inv(0)` / `det(I)` short-circuits), or
- Soften the prose to "the fuzz generator avoids producing `inv(skew)` over odd dimensions" and note the library does not reject these.

Issue #89 has the same wording and should be updated alongside.

## Correctness issues

**`include/numsim_cas/tensor/visitors/tensor_evaluator.h:367-372`** — unguarded `std::any_cast<ValueType>(val)` will throw `bad_any_cast` whenever the parent `m_scalar_eval` stores values at a different precision than `ValueType` (e.g. parent `double`, evaluator typed on `float`). There's no `if (val.type() == typeid(ValueType))` filter. The same loop also `static_pointer_cast`s every key to `scalar_expression` regardless of its actual underlying type — if any key isn't a scalar expression, that's UB the first time `t2s_eval` reads it. Filter by `id()` before forwarding.

**`include/numsim_cas/functions.h:33-42` and `include/numsim_cas/core/simplifier/simplifier_add.h:298-307`** — both resolve only one level of collision. If `combined = existing + new` itself simplifies to something whose key collides with yet another entry in the map, `push_back` throws the same `internal_error` this fix was added to prevent. Needs to loop until clean. Both blocks also implement the same algorithm; should be one helper on `n_ary_tree` (e.g. `merge_or_insert`) so they can't drift.

The comments in both blocks (`x+x=2*x collides with existing 2*x`) are wrong: per the n_ary_tree hash design (coefficient not part of the key), `x` and `2*x` map to the **same** key and cannot coexist. The real collision is when `+` algebraically simplifies to a *different* key (`cos(x)^2 + sin(x)^2 → 1`, `x + (-x) → 0`, projector identities). Worth fixing the comment so the next reader doesn't model the wrong failure mode.

**`include/numsim_cas/scalar/visitors/scalar_differentiation.h:308-318`** — the reset of `m_result` on entry is the right fix, but returning `get_scalar_zero()` when the visitor doesn't write `m_result` silently swallows the "missing override for a new node type" case. A node added to the visitor base without an `operator()` here will now produce zero derivatives instead of a loud failure. Consider an assert/throw when `expr.is_valid()` but `m_result` is still empty post-`accept()`.

## Design smells

**`include/numsim_cas/tensor/tensor_operators.h:60-83`** — commit `7e962e5`'s message says the Skew space annotation "is not preserved on all platforms when stored inside `tensor_mul`." If the annotation is platform-dependent, every callsite that reads `data()->space()` is suspect, and this PR's strategy of annotating `trans(A) - A` with a Skew space is fragile by construction. Worth identifying the root cause (likely `symbol_map` iteration order across stdlib implementations) and fixing it at the source rather than working around it in both the operator and the test helper.

**`tests/FuzzyTensorDiffTest.h:38-74`** — the projector-based skew detection duplicates structural classification that the projector simplifier already knows about. If projector logic evolves, this test helper will silently drift. Better to expose the classification from `tensor/` and have the test call into it.

## Missing regressions

All six bugs are demonstrated only via re-enabling fuzz seeds. None has a deterministic unit test:

- merge_add collision path
- `scalar_pow` differentiation with `g` non-constant and `h` constant (or vice versa)
- `tensor_to_scalar_with_tensor_mul` scalar forwarding
- `is_same` on invalid expression

Future regressions will surface as "seed 47 is flaky again," which is harder to bisect than a named test. A small handful of targeted tests would lock in the contracts the fuzz suite is currently the only witness for.

petlenz:
Pushed `f76bb7f` addressing the correctness items:

- **`tensor_evaluator.h` forwarding loop** — now filters by `std::any::type()` against `typeid(ValueType)` and uses `dynamic_pointer_cast<scalar_expression>` to keep only genuine scalar keys. The unguarded `any_cast`, force `static_pointer_cast`, and unnecessary `const_pointer_cast` are all gone.
- **Duplicate-insertion handling** — extracted into `n_ary_tree::merge_or_insert(entry)` which loops until the result has no further collisions. Both `merge_add` (`functions.h`) and the `n_ary_add_dispatch` `SymbolType` overload (`simplifier_add.h`) now call the single helper. Misleading `x+x=2*x collides with 2*x` comment replaced with one describing the actual failure mode (algebraic simplification producing a different key).

The `apply_imp` item from my review I'm withdrawing — scalar visitors are pure-virtual (`visitor_const<T>::operator()(T const&) = 0`), so a missing override is a compile error, not a runtime silent-zero. The empty-`m_result`-treated-as-zero pattern is intentional and used by `scalar_negative` / `scalar_pow` / etc., so the current code is correct.

The remaining review points (library-level `inv(skew_odd)` rejection vs. test-side filter, root cause of platform-dependent Skew space, deterministic regressions for each bug) are design/scope items, not correctness — leaving those open.

All 1482 tests pass.

petlenz:
Pushed `1b387f4` addressing the design issues raised earlier.

- **Library-level `inv(skew_odd)` rejection.** New `include/numsim_cas/tensor/skew_classification.h` exposes `is_provably_skew()` / `contains_skew_factor()` as library API. `inv()` in `tensor_functions.h` now throws `invalid_expression_error` at construction when the operand is rank-2, odd-dim, and contains a provably skew factor — same construction-time rejection pattern as the existing `inv(I)` / `inv(kronecker)` short-circuits. Hand-written `inv(skew(A))` no longer produces silent NaN at evaluation.
- **No more duplication between test and library.** The fuzz test now `try`/`catch`es around `inv()` against the library API instead of carrying its own structural detector. The leftover `[INV-CHECK]` / `[REJECT-INV]` diagnostic `cerr` output goes with it.
- **Deterministic regressions.** Added to `CoreBugFixTest`:
  - `IsSameOnInvalidExpressionReturnsFalse`
  - `ScalarPowDiffConstBaseVariableExponent` — `d/dx(2^x) = 2^x · log(2)`, the new branch in `scalar_differentiation.cpp`
  - `ScalarPowDiffVariableBaseConstExponent`, `ScalarPowDiffBothConstWrtArg`
  - `AddMergeHandlesSimplificationCollision` — exercises `merge_or_insert` against a `cos²+sin²→1` collapse
- **`TensorSpacePropagationTest`.** Replaced `InvSkewPreservesSkew` (which asserted a mathematically nonsensical property in `dim=3`) with `InvSkewRejectedInOddDim` plus `InvSkewPreservesSkewInEvenDim` (using a locally constructed `dim=4` `W4`).
- **PR description updated** to match the new reality.

What's still left from the original review:

- **Root cause of platform-dependent Skew space annotation.** Commit `7e962e5`'s observation that the annotation isn't preserved through `tensor_mul` is real, but tracking it down requires a wider audit of how `tensor_space` propagates across rewrite paths — and the library now does structural matching as the authoritative check, so the annotation is a fast path rather than a contract. Worth a separate issue if we want to make the annotation reliable everywhere.

1488 tests pass.

petlenz:
Second pass after the design fixes — couple of real bugs introduced or missed, plus cleanup items.

## Correctness

**`include/numsim_cas/tensor/skew_classification.h:78-87`** — `contains_skew_factor` walks `tensor_mul.data()` and `inner_product_wrapper.expr_lhs()/expr_rhs()` but only calls `is_provably_skew(child)` on each, not `contains_skew_factor(child)`. That means a `tensor_mul` whose child is a `tensor_scalar_mul(c, skew(A))` (or a `tensor_negative(skew(A))`, or a nested `tensor_mul`) is **not** caught — `is_provably_skew` doesn't peek through scalar multiples or negation. So `inv(B * (2*skew(A)))` over an odd dim won't be rejected even though it's still singular. The `tensor_scalar_mul` / `tensor_negative` branches further down do recurse correctly — the bug is just that the `tensor_mul` and `inner_product` walks short-circuit too early.

Fix: have those two loops call `contains_skew_factor` on each child instead of `is_provably_skew`.

**`include/numsim_cas/tensor/skew_classification.h:31-32`, `:74-75`** — both functions are declared `noexcept`, but they call `e.get()` (throws `invalid_expression_error` on invalid input), `e.get().space()`, and `e.template get<T>()`. If a caller ever passes an invalid `expression_holder`, the throw inside a `noexcept` becomes `std::terminate`. The actual call from `inv()` always has a valid `expr`, but the public API of a library header shouldn't make this kind of contract on its caller without enforcing it. Either drop `noexcept` or add an `if (!e.is_valid()) return false;` guard at the top of each.

## Test quality

**`tests/CoreBugFixTest.h:244-255` (`AddMergeHandlesSimplificationCollision`)** — this test doesn't exercise the transitive-collision path it claims to. There's no `sin²(x) + cos²(x) → 1` simplification rule in the scalar simplifier, so `pow(cos(x), 2) + pow(sin(x), 2)` stays as a two-term `scalar_add` and never collides with the constant `2`. The assertion `EXPECT_NO_THROW` passes trivially for any expression construction.

The fuzz suite is currently the only test that actually hits the transitive collision. To get a deterministic regression, the minimal repro from one of the previously-skipped seeds should be extracted — or admit this is a smoke test only and rename it accordingly.

**`tests/FuzzyDiffBase.h:221-280`** — leftover diagnostic `[SKIP]` `std::cerr` output left over from the debugging cycle. Running the fuzz suite now produces:

```
[SKIP] seed=23 phase=verification var=E exception=make_tensor_data_imp::evaluate(dim, rank) rank > MaxRank || rank == 0
[SKIP] seed=63 phase=verification var=F exception=...
[SKIP] seed=10046 phase=verification var=C exception=...
```

9 lines per fuzz run today, but the same instrumentation will spew on every future failing seed. Either remove (the catch arms already return `Skip` / route through `handle_exception`, which is the proper signaling channel) or gate behind a `NUMSIM_CAS_VERBOSE_FUZZ` env var.

## Minor

**`tests/FuzzyTensorDiffTest.h:8`** — `#include <numsim_cas/tensor/skew_classification.h>` is now unused after the refactor; the test no longer references `is_provably_skew` / `contains_skew_factor`. Dead include.

**`include/numsim_cas/tensor/tensor_functions.h:378-382`** —
1. The rank check is hard-coded to 2. Rank-4 inverses of skew projectors over odd dim aren't rejected. Probably intentional (rank-4 inv is rare and tmech::inv handling differs), but a one-line comment saying "rank-2 only; rank-4 inv() is a separate concern" would help the next reader.
2. The error message says `"skew-symmetric tensor in odd dimensions is singular"` but `contains_skew_factor` also rejects expressions that *contain* a skew factor without being themselves skew. E.g. `inv(B * skew(A))` produces the same message even though `B * skew(A)` isn't skew, just singular. Saying `"inv: operand contains a singular skew factor in odd dimensions"` matches the actual check.

## Not blocking

- `merge_or_insert` termination is bounded by `m_symbol_map.size()` (each iteration erases one entry), so unbounded looping is impossible. ✓
- `dynamic_pointer_cast<scalar_expression>` requires the polymorphic base; verified `expression` has virtual functions. ✓

petlenz:
Pushed `c82bffe` for the second-pass items.

- **`contains_skew_factor` now recurses** into `tensor_mul` children and both `inner_product_wrapper` sides via `contains_skew_factor` instead of `is_provably_skew`. On this build the space annotation propagates through `tensor_scalar_mul` / `tensor_negative` so the fast-path catches scaled/negated skew already — the recursion is defense-in-depth for paths where annotation is lost (the original motivation for structural detection in the first place).
- **Dropped `noexcept`** from `is_provably_skew` / `contains_skew_factor` and added `is_valid()` guards. The internals call `e.get()` which throws on an invalid holder, and a `noexcept` body that throws is `std::terminate`. The current single caller in `inv()` always passes a valid `expr`, but the public library header shouldn't depend on that.
- **`inv()` rejection** — error message now says "operand contains a skew-symmetric factor in odd dimensions (singular)" to match what `contains_skew_factor` actually checks. Added a code comment that the rank check is rank-2 only and rank-4 inverses are a separate concern.
- **`FuzzyDiffBase` `[SKIP]` cerr noise removed** — the catch arms already signal `TestResult::Skip` through the return path; the prints were just log spam from the debugging cycle. `0` `[SKIP]` lines now on a fuzz run (vs. 9 before).
- **Dead include removed** from `tests/FuzzyTensorDiffTest.h` (`skew_classification.h` was no longer referenced after the earlier refactor).
- **Renamed `AddMergeHandlesSimplificationCollision`** to `AddCompoundConstructionSmoke` with an honest comment — the codebase has no `sin²+cos²→1` rule so the original test didn't exercise the transitive-collision path. The fuzz suite still covers the real path; the renamed test is a construction smoke check.
- **Added `InvRejectsScaledSkewFactorInTensorMul` / `InvRejectsNegatedSkewFactorInTensorMul`** as contract documentation. They pass via the space-annotation fast path on this build but lock in the expected rejection so future annotation-propagation regressions don't quietly disable the recursion path.

1490 tests pass.

petlenz:
Third pass. Most concerns addressed; a couple of smaller things left.

## Worth a closer look

**`include/numsim_cas/tensor/tensor_operators.h:63`** — the `dim() % 2 != 0` guard on the `trans(A) − A → Skew` annotation couples *detection* with the *singularity reason*. The expression `trans(A) − A` is mathematically skew-symmetric in **any** dimension (its definition is `(trans(A)−A)^T = −(trans(A)−A)`); only its singularity under `inv()` is odd-dim specific. By gating the annotation on odd dim, an even-dim construction like `dim=4, trans(A) − A` does **not** carry a Skew annotation, so downstream patterns that depend on it (e.g. `skew(trans(A) − A) → trans(A) − A`, or any projector simplification keyed on Skew) won't fire.

Cleaner split: always annotate `trans(A) − A` with Skew at construction, and leave the `dim % 2 != 0` check where it belongs — in `inv()`'s `contains_skew_factor` rejection. That decouples "what this expression *is*" from "what cannot be inverted."

## Smaller

**`include/numsim_cas/core/evaluator_base.h:24` — `symbol_values()` returns a `const&` to the protected `m_symbols_to_value` map**, promoting an implementation-detail member to a public read-anyone API. Single consumer is `tensor_evaluator`'s `tensor_to_scalar_with_tensor_mul` forwarding loop. Cleaner alternatives:
- A `forward_scalars_to(other_evaluator&)` method that the parent calls — keeps the map private and pushes the filtering logic to one place.
- Friend access for `tensor_evaluator`.

Either keeps the map encapsulated; the current getter lets any future caller iterate the map and do unsafe things with the `std::any` values.

**`include/numsim_cas/tensor/tensor_operators.h:78`** — `result.data()->set_space({Skew{}, AnyTraceTag{}})` mutates the underlying expression node *after* `accept(visitor)` has constructed and (potentially, on first hash request) hashed it. The hash design excludes the space annotation, so this works today — but it's an implicit dependency. If anyone ever folds space into the hash, this silently produces an expression whose stored hash disagrees with its space. Worth a one-line comment marking the dependence, or a `set_space_before_finalize()` helper that asserts no hash has been computed.

**`include/numsim_cas/core/n_ary_tree.h:61-71` — `merge_or_insert` is only weakly exception-safe across iterations.** If `+` throws on the Nth iteration (N>1), prior erasures are committed: the map has lost entries it had at entry. Current callers (`merge_add`, `n_ary_add_dispatch`) discard the result on exception, so this doesn't manifest as a visible bug, but a one-line comment that the method is **not** atomic under exception would prevent surprises if someone reuses it from a context that catches and continues.

## Confirmed good

- `contains_skew_factor` recursion ✓
- `is_valid()` guards + `noexcept` removed from the skew helpers ✓
- `inv()` error message and rank-2 comment ✓
- `[SKIP]` cerr cleanup ✓
- Dead include removed ✓
- Honest renaming of the merge smoke test ✓

Pre-existing observations from earlier reviews that remain open (call out for a follow-up PR rather than block this one):
- Deterministic regression for `merge_or_insert`'s transitive-collision loop (the fuzz suite is the only exercise).
- Root cause of platform-dependent Skew-space loss inside `tensor_mul`.
- Performance of re-building `t2s_eval` per `tensor_to_scalar_with_tensor_mul` visit.

petlenz:
Pushed `972697c` for the third-pass items.

- **Decoupled the Skew annotation from the dim guard.** `tensor_operators` now annotates `trans(A) − A` and `A − trans(A)` as Skew unconditionally on rank-2; `inv()` keeps the `dim % 2 != 0` check for the singularity rejection. Added `TransMinusSelfIsAnnotatedSkewInEvenDim` — confirmed to fail before the fix (`is_skew(trans(A4) − A4) == false` under the old guard) and pass after.
- **Encapsulated symbol forwarding.** Removed `evaluator_base::symbol_values()` (which leaked the protected map). Added `scalar_evaluator::forward_values_to(target)` that filters by `std::any::type()` and `dynamic_pointer_cast<scalar_expression>` internally. `tensor_evaluator`'s `tensor_to_scalar_with_tensor_mul` forwarder is now a single call: `m_scalar_eval.forward_values_to(t2s_eval)`.
- **Documented `merge_or_insert`'s exception-safety contract.** Comment notes that the method is not atomic — if `+` throws on iteration `N>1`, prior erasures are committed. Current callers discard on throw, so no visible bug; the comment prevents future misuse.
- **Documented the `set_space` post-construction dependency.** One-line note that the mutation is safe because the n_ary_tree hash excludes the space annotation, and that this must move if hash coverage ever changes.

1491 tests pass (one new deterministic regression).

petlenz:
Fourth pass. PR is in good shape; one minor symmetry observation and one small testing gap, neither blocking.

## Minor

**`include/numsim_cas/tensor/tensor_operators.h:24-45` (`add` branch) vs `:52-86` (`sub` branch)** — asymmetry in skew annotation. The `sub` operator now annotates `trans(A) − A` and `A − trans(A)` as Skew at construction. The `add` operator does **not** apply the same treatment to `trans(A) + (−A)` / `(−A) + trans(A)`, which is the same expression up to sign distribution and equally skew-symmetric by definition.

`contains_skew_factor` in `skew_classification.h` catches the `add(trans, neg)` form structurally, so `inv()` rejection works for both. The miss is for non-`inv()` consumers that fast-path on the Skew annotation — e.g. `skew(trans(A) + (-A))` won't simplify via the projector identity the way `skew(trans(A) − A)` will. Probably uncommon in user code (people write `−` rather than `+ (−)`), but worth tracking as a follow-up for full coverage.

**`forward_values_to`'s filter paths are untested.** Tests exercise the happy path (scalar keys → scalar evaluator). The `val.type() != typeid(ValueType)` continue and the `dynamic_pointer_cast` returning null continue are defensive guards with no coverage. Not strictly required since they're defense against a "shouldn't happen" condition, but a small targeted test would lock in the behavior — e.g. `set` a tensor symbol on the scalar evaluator (via the templated `set<ExprBase>` overload), then verify `forward_values_to` skips it without throwing.

## Confirmed good

- Skew annotation now dim-agnostic; `inv()` keeps the singularity check ✓
- `evaluator_base::symbol_values()` getter removed; encapsulation restored ✓
- `forward_values_to` filters internally, single-line call site ✓
- `merge_or_insert` exception-safety contract documented ✓
- `set_space` post-construction hash dependency documented ✓
- `TransMinusSelfIsAnnotatedSkewInEvenDim` regression verified to fail-before / pass-after ✓

## Outstanding (cross-PR follow-ups, not blocking)

- Deterministic regression for `merge_or_insert`'s transitive collision loop (still only fuzz-covered).
- Root cause of platform-dependent Skew-space loss inside `tensor_mul`.
- Performance: rebuilding `t2s_eval` per `tensor_to_scalar_with_tensor_mul` visit.

Approving the direction; the two minor items above can land in this PR or follow-up — your call.

petlenz:
Pushed `8a7c66d` for the fourth-pass items.

- **`add` branch symmetric with `sub`.** `trans(A) + (-A)` and `(-A) + trans(A)` now get the same Skew annotation that `trans(A) - A` does. Same lambda pattern as the sub branch. Added `TransPlusNegIsAnnotatedSkew` — confirmed to fail before the fix and pass after.
- **`forward_values_to` non-scalar-key skip path tested.** `ForwardValuesToSkipsNonScalarKey` plants a tensor symbol into a `scalar_evaluator` (via the templated `set<ExprBase>`) and verifies `forward_values_to(tensor_evaluator)` survives without UB. Locks in the `dynamic_pointer_cast<scalar_expression>` filter contract.

1493 tests pass.

petlenz:
Fifth pass. The add-side merge collision fix never made it across to the **sub-side** dispatchers in `simplifier_sub.h` — same bug class, three call sites still using the single-shot `find / erase / push_back` pattern that originally tripped the duplicate-child `internal_error` in `merge_add`.

## Real finding

**`include/numsim_cas/core/simplifier/simplifier_sub.h:96-109, :218-241, :243-258` — three sites still vulnerable to the transitive-collision bug.**

The PR moved `merge_add` (functions.h) and the `n_ary_add_dispatch::dispatch<SymbolType>` (simplifier_add.h) onto `n_ary_tree::merge_or_insert`. The symmetric sub-side dispatchers were left on the old pattern:

1. **`:96-109`** — `dispatch(typename Traits::add_type const &rhs)` for `-expr - (coeff + x) -> -(expr + coeff + x)`:
   ```cpp
   auto pos = add.symbol_map().find(inner);
   if (pos != add.symbol_map().end()) {
     auto combined = pos->second + inner;
     add.symbol_map().erase(pos);
     add.push_back(combined);  // <-- collision risk if `+` simplifies to a different key
   }
   ```

2. **`:218-241`** — `dispatch(typename Traits::add_type const &rhs)` for "merge two add expressions in subtraction". The child-child combination case uses `add.push_back(child + pos->second)` — same pattern as the original buggy `merge_add`.

3. **`:243-258`** — `dispatch(SymbolType)` for `x + y + z - x -> y + z`:
   ```cpp
   auto expr{pos->second - base::m_rhs};
   add.symbol_map().erase(pos);
   add.push_back(expr);  // <-- collision risk if `-` simplifies to a different key
   ```

Each site has the same failure mode: if the combination algebraically simplifies to an expression whose key now matches a *different* existing entry in the result map, `push_back` throws `internal_error` — exactly the bug that `merge_or_insert`'s loop was added to prevent.

Worth converting all three to `merge_or_insert` for consistency (and to close the bug-class instead of patching only the add side). The change is mechanical and the loop already handles the `+` vs `-` combine since the operator on the holders is just whatever is in scope at the call site (for site 3, the loop's internal `+ entry` would need to match what the caller does — so either generalize `merge_or_insert` to take a combiner, or have sub-side sites just push `-m_rhs` style entries directly and let `+` collapse them).

A simpler API: `merge_or_insert(entry)` always uses `+` internally, which is exactly right for any site that wants "add this entry, combining with any existing collision." The sub-side just needs to pass the negated value where appropriate (`merge_or_insert(-m_rhs)` instead of `push_back(-m_rhs)`).

## Minor

**`include/numsim_cas/tensor/tensor_operators.h:46-55` and `:94-101` — parallel-but-not-identical detection logic for `trans(A) + (−A)` and `trans(A) − A`.**

The add branch uses a `matches_trans_plus_neg` lambda; the sub branch uses inline bool expressions checking `bc.expr() == rhs` and `bc.expr() == lhs` (no negation on the rhs side, since sub doesn't wrap in negative). Functionally correct, but the two helpers are similar enough to invite drift. A single helper:

```cpp
// True iff `a` is trans(X) where X is the operand that, combined with `b`,
// yields a skew expression.
template <typename Holder>
bool is_trans_of(Holder const& a, Holder const& target);
```

…called from both branches would keep the two patterns in lockstep.

## Confirmed good

- `add` branch now annotates `trans(A) + (-A)` symmetric with `sub` ✓
- `ForwardValuesToSkipsNonScalarKey` test added ✓
- All earlier fixes hold up ✓

## Open for follow-up (still not blocking)

- Deterministic regression for `merge_or_insert`'s transitive loop (still fuzz-only).
- Root cause of platform-dependent Skew-space loss in `tensor_mul`.
- `t2s_eval` per-visit rebuild.

petlenz:
Pushed `bf75393` for the fifth-pass items.

- **Three sub-side dispatchers in `simplifier_sub.h` now use `merge_or_insert`:**
  - `negative_sub_dispatch::dispatch(add_type)` (`-inner - (coeff + ...)`)
  - `n_ary_sub_dispatch::dispatch(add_type)` (merge two add expressions)
  - `n_ary_sub_dispatch::dispatch(SymbolType)` (`x+y+z - x`)

  Where the dispatch combines via `-` (the symbol-rhs case), I kept the explicit `find / combine via -` step and routed the final push through `merge_or_insert` — that preserves the existing `pos->second - m_rhs` semantic and still closes the transitive-collision bug class.

- **Shared `is_trans_of(a, target)` helper** in `tensor_operators.h`. The add branch's `is_trans_of_neg` lambda now thin-wraps it (just unwraps `tensor_negative`); the sub branch uses it directly. Two patterns can't drift any more.

- **Two new smoke tests in `CoreBugFixTest`:** `SubSymbolDispatchSmoke` and `NegativeSubAddDispatchSmoke`. A third smoke for the merge-two-adds dispatch would surface a separate, pre-existing latent bug in that dispatch (unguarded `lhs.coeff() + rhs.coeff()` against invalid coefficients, plus `+` instead of `-` for subtraction). Out of scope here; documented inline as a marker rather than adding a failing test.

1495 tests pass.

petlenz:

After multiple rounds of self-review (three correctness items, two design items, scope rewrite), the PR is in good shape. 1495 tests pass at the tip. Six follow-on PRs (#98, #100, #103, #105, #106, #107, #108) are stacked and reviewed.

No new findings. Ready to merge from my side.

Issue #97, opened 2026-05-15
Add n_ary_mul_mul_dispatch (LHS=mul) sibling for the mul dispatcher family


petlenz:
## Summary

The mul family of simplifier dispatchers in `include/numsim_cas/core/simplifier/simplifier_mul.h` has only the base `mul_dispatch` class. There is no LHS-specialised sibling for the case where the left operand is itself a `mul` expression.

Compare with the add family in `simplifier_add.h`, which has dedicated dispatchers for every LHS shape (`constant_add_dispatch`, `one_add_dispatch`, `n_ary_add_dispatch`, `n_ary_mul_add_dispatch`, `symbol_add_dispatch`, `negative_add_dispatch`). The sub family has the same parallel structure.

Surfaced during the simplifier coverage audit on branch `95-unify-simplifier-dispatchers` (issue #95).

## What's missing

Rules that need a `n_ary_mul_mul_dispatch<Traits>` to fire on the LHS=mul side:

- **Coefficient combining**: `(c₁·x·y) · (c₂·x·z) → c₁·c₂·pow(x,2)·y·z`. Today the coefficients aren't extracted before the merge — they stay on the children, and the resulting mul has redundant constants.
- **Like-factor combining via `merge_or_insert`-style logic on `n_ary_vector`**: when LHS and RHS both have a child with the same key, combine them. Currently happens only via the per-domain wrapper `n_ary_mul::dispatch(mul_type)` in each domain, with inconsistent coverage.
- **Pow promotion**: `(x·y) · (x·z) → pow(x,2)·y·z`. Today done in the per-domain `symbol_mul` for the `x · x → pow(x,2)` case but not for the n-ary version.

## Related

This is the mul-side equivalent of what `n_ary_add_dispatch` + the `merge_or_insert` helper do for add. Issue #17 ("merge two n_ary_trees") asks for the same thing from the user side; this issue is the dispatcher-layer specification.

`n_ary_vector` (the mul backing store) has no equivalent of `n_ary_tree::merge_or_insert`. That helper would need to be added or substituted by a domain-specific equivalent.

## Acceptance criteria

- [ ] Add `n_ary_mul_mul_dispatch<Traits>` in `include/numsim_cas/core/simplifier/simplifier_mul.h` paralleling `n_ary_add_dispatch` in `simplifier_add.h`.
[...]

Issue #96, opened 2026-05-15
Add tensor pow simplifier (missing entire family)


petlenz:
## Summary

The tensor domain is missing a `pow` simplifier entirely. Compare with the scalar and tensor-to-scalar domains, both of which have a full `simplifier_pow` family that hooks into the generic `pow_dispatch` / `pow_pow_dispatch` / `mul_pow_dispatch` machinery in `core/simplifier/simplifier_pow.h`.

Surfaced during the simplifier coverage audit on branch `95-unify-simplifier-dispatchers` (issue #95).

## What's missing

- No `include/numsim_cas/tensor/simplifier/tensor_simplifier_pow.h` file.
- Tensor `pow` expressions are constructed via `include/numsim_cas/tensor/functions/tensor_pow.h` and the `pow()` free function, but no simplifier visitor runs on them at construction time.

Rules that exist in scalar `pow` but are missing for tensor:

- `pow(pow(A, m), n) → pow(A, m·n)` (the `pow_pow_dispatch` path).
- `pow(mul, -rhs) → 1 / mul` extraction / cancellation (the `mul_pow_dispatch` path).
- `pow(x, 0) → 1`, `pow(x, 1) → x` (handled inline in some constructors but not consistently).
- `pow(inv(A), n) → inv(pow(A, n))` and the inverse pair.

## Why no one's hit it hard

Tensor `pow` is currently used mostly with constant integer exponents in
expressions like `pow(C, 2)` and `pow(C, -1)`. The construction-time path in
`tensor_pow.h` handles the `n = 0` and `n = 1` short-circuits inline. Compound
cases (pow of pow, inverse-pow, mul-pow distribution) just produce a
symbolic-but-not-simplified node, which still evaluates correctly via the
runtime tensor_data_pow path.

It becomes more visible when:
- Differentiation produces nested `pow(pow(...), ...)` chains that don't collapse.
- Evaluator is called on a `pow(inv(A), n)` shape that should evaluate as `pow(A, -n)` for half the cost.

## Acceptance criteria

- [ ] Add `include/numsim_cas/tensor/simplifier/tensor_simplifier_pow.h` with wrapper classes that inherit from the generic dispatchers in `core/simplifier/simplifier_pow.h` (same pattern as `scalar/simplifier/scalar_simplifier_pow.h`).
- [ ] Wire it into the `pow()` free function for tensors so the visitor runs at construction time (mirror what `pow()` does in `scalar/scalar_std.h`).
[...]

Issue #95, opened 2026-05-15
Unify simplifier dispatchers across add/sub/mul/pow and across domains


petlenz:
## Summary

The simplifier dispatchers in `include/numsim_cas/core/simplifier/` and the per-domain wrappers in `scalar/simplifier/`, `tensor/simplifier/`, `tensor_to_scalar/simplifier/` have grown in parallel but lack feature parity. This shows up as missing simplifications on some sides, accidental copy-paste-without-converting bugs, and inconsistent collision-handling patterns across the add / sub / mul / pow axes.

PR #90 exposed several of these (and fixed some). Time for a deliberate parity audit.

### add vs sub on the same domain

- **Add side** uses `n_ary_tree::merge_or_insert` everywhere relevant (fixed in PR #90).
- **Sub side** was partially converted in PR #90 (commit `bf75393`), but the `n_ary_sub_dispatch::dispatch(add, add)` branch still has a pre-existing semantic bug — uses `+` where `-` would be correct, and unguarded coeff addition. Tracked in #91.
- The `trans(A) − A → Skew` annotation that lives in the sub operator's branch has a mirror in the add operator (added in PR #90) — they used to drift. The shared `is_trans_of` helper now keeps them in lockstep, but no equivalent shared helper exists for other paired add/sub branches.

### scalar vs tensor vs t2s

The generic dispatcher templates in `core/simplifier/` are domain-agnostic, but each domain has its own thin wrapper class that exposes the dispatch methods. The wrappers are mostly mechanical (CRTP / dispatch macros) but small differences accumulate:

- `tensor_to_scalar` simplifier wrappers declare virtual functions in `.h` and define in `.cpp` (to keep `pow()` ADL resolution in the right TU). Scalar wrappers are header-only.
- Some simplification rules exist only in one domain. Examples to audit:
  - The scalar pow simplifier handles many constant-folding cases that the tensor t2s pow doesn't.
  - `tensor_scalar_mul` propagates `space()` from its rhs (good); equivalent t2s wrappers don't have an analogous property to propagate.

### mul vs add for n-ary collisions

PR #90 closed the add-side transitive-collision bug class via `n_ary_tree::merge_or_insert`. The **mul** side uses `n_ary_vector` (a different backing store) and has no equivalent helper. Issue #17 calls out the same `(x*y*z) * (x*a) → pow(x,2)*y*z*a` simplification but the underlying collision-handling pattern hasn't been audited the way the add side was.

[...]

Issue #94, opened 2026-05-15
tensor_evaluator rebuilds t2s_eval per tensor_to_scalar_with_tensor_mul visit


petlenz:
## Summary

`tensor_evaluator<ValueType>::operator()(tensor_to_scalar_with_tensor_mul const &)` constructs a fresh `tensor_to_scalar_evaluator<ValueType>` *every time* it visits a `tensor_to_scalar_with_tensor_mul` node, and copies all tensor values plus filters/copies all scalar values into it via `m_scalar_eval.forward_values_to(t2s_eval)`.

An expression with many `tensor_to_scalar_with_tensor_mul` subnodes pays this cost on every visit during a single evaluation: `O(visits × (|tensor_values| + |scalar_values|))`.

## Where

`include/numsim_cas/tensor/visitors/tensor_evaluator.h` — the out-of-line `operator()(tensor_to_scalar_with_tensor_mul const &)` definition near the bottom of the file (introduced/cleaned up across commits `6d2fbc2`, `f76bb7f`, and `972697c` on PR #90).

```cpp
template <typename ValueType>
void tensor_evaluator<ValueType>::operator()(
    tensor_to_scalar_with_tensor_mul const &visitable) {
  tensor_to_scalar_evaluator<ValueType> t2s_eval;           // <-- fresh per visit
  for (auto const &[key, val] : m_tensor_values) {          // <-- O(|tensors|) per visit
    t2s_eval.set(key, val);
  }
  m_scalar_eval.forward_values_to(t2s_eval);                // <-- O(|scalars|) per visit
  const auto scalar_val = t2s_eval.apply(visitable.expr_rhs());
  ...
}
```

## Why this is wasteful

The tensor and scalar value maps don't change during a single `tensor_evaluator::apply()` call. We re-copy them N times where N is the number of `tensor_to_scalar_with_tensor_mul` nodes in the expression tree. For typical mechanics expressions with a handful of `f(A) * B + g(A) * C + h(A) * D` terms, that's 3× redundant copies; for more complex expressions, more.

This is the only out-of-line evaluator method that pays this cost, because `tensor_to_scalar_with_tensor_mul` is the only one that needs to build a `tensor_to_scalar_evaluator` to evaluate its RHS.

## Options

[...]

PR #98, opened 2026-05-15
Audit simplifier dispatchers and fix n_ary_sub_dispatch(add, add) bug

### LINKED ISSUE #91: n_ary_sub_dispatch::dispatch(add, add) has unguarded coeff add and uses + instead of -
## Summary

`n_ary_sub_dispatch::dispatch(typename Traits::add_type const &rhs)` — the "merge two add expressions" branch in `include/numsim_cas/core/simplifier/simplifier_sub.h` — has two pre-existing defects that surface when both operands are `add` expressions with the same coefficient state.

Discovered during the PR #90 review while routing the other sub-side dispatchers through `n_ary_tree::merge_or_insert`. Out of scope for that PR; filing here for follow-up.

## Repro

```cpp
auto [x, y, z] = make_scalar_variable("x", "y", "z");
auto lhs = x + y + z;       // scalar_add, no constant coefficient
auto rhs = x + y;           // scalar_add, no constant coefficient
auto diff = lhs - rhs;      // throws invalid_expression_error
```

`(x+y+z) - (x+y)` should simplify to `z`. Instead it throws `"expression_holder: access to invalid (null) expression"`.

## Root cause

The dispatch body, abbreviated:

```cpp
  auto expr{make_expression<typename Traits::add_type>()};
  auto &add{expr.template get<typename Traits::add_type>()};
  add.set_coeff(lhs.coeff() + rhs.coeff());          // <-- defect 1 and 2
  ...
  for (auto &child : lhs.symbol_map() | std::views::values) {
    auto pos{rhs.symbol_map().find(child)};
    if (pos != rhs.symbol_map().end()) {
      add.merge_or_insert(child + pos->second);      // <-- defect 2 (also here)
    } else {
      add.merge_or_insert(child);
    }
  }
  ...
}
```

### Defect 1 — unguarded coefficient addition

`lhs.coeff() + rhs.coeff()` is only safe when both holders are valid. When both adds have no constant term (their default state after `x + y + z` style construction), both coeffs are default-constructed (invalid). `tag_invoke(add_fn, invalid, invalid)` calls `lhs.get<scalar_visitable_t>()` which throws `invalid_expression_error`.

The symmetric routine in `functions.h` already handles this correctly:

```cpp
// merge_add — already robust
if (lhs.coeff().is_valid() && rhs.coeff().is_valid()) {
  result.set_coeff(lhs.coeff() + rhs.coeff());
} else {
  if (lhs.coeff().is_valid()) result.set_coeff(lhs.coeff());
  if (rhs.coeff().is_valid()) result.set_coeff(rhs.coeff());
}
```

The sub dispatch should use the same shape (with `-` semantics — see defect 2).

### Defect 2 — wrong sign for subtraction

The function lives in `simplifier_sub.h` and constructs the result of `lhs_add - rhs_add`. Mathematically:

```
(c_l + a + b) - (c_r + a + d) = (c_l - c_r) + 0 + b - d
                              = (c_l - c_r) + b - d
```

- Coefficient should be `lhs.coeff() - rhs.coeff()`, not `+`.
- Child-child combination (when both add children share a key) should be `child - pos->second`, not `child + pos->second`.
- Children present only in `rhs` should be added as `-child`, not `child`.

The current implementation looks like a copy-paste from the corresponding **add** routine that was never converted to subtraction semantics.

## Why existing tests don't catch this

Two factors:
1. The combination of (a) both LHS and RHS being `add` types and (b) both having invalid coefficients isn't exercised by any current named test. The fuzz suite may have hit it intermittently but `cas_error` is caught and the seed silently skipped.
2. The `+`-instead-of-`-` defect produces a wrong but non-throwing result on most inputs (e.g. `(2+x) - (1+x)` would compute `3 + 2x` instead of `1`), but no test asserts the value of a 2-add subtraction whose result depends on the merge path.

## Fix sketch

```cpp
  auto expr{make_expression<typename Traits::add_type>()};
  auto &add{expr.template get<typename Traits::add_type>()};

  // Coefficient: c_l - c_r, guarded against invalid sides.
  if (lhs.coeff().is_valid() && rhs.coeff().is_valid()) {
    add.set_coeff(lhs.coeff() - rhs.coeff());
  } else if (lhs.coeff().is_valid()) {
    add.set_coeff(lhs.coeff());
  } else if (rhs.coeff().is_valid()) {
    add.set_coeff(-rhs.coeff());
  }

  std::set<expr_holder_t> used_expr;
  for (auto &child : lhs.symbol_map() | std::views::values) {
    auto pos{rhs.symbol_map().find(child)};
    if (pos != rhs.symbol_map().end()) {
      used_expr.insert(pos->second);
      add.merge_or_insert(child - pos->second);   // was `+`
    } else {
      add.merge_or_insert(child);
    }
  }
  if (used_expr.size() != rhs.size()) {
      if (!used_expr.count(child)) {
        add.merge_or_insert(-child);              // was `child`
      }
    }
  }
  return expr;
}
```

## Acceptance criteria

- [ ] `(x+y+z) - (x+y)` simplifies to `z` without throwing.
- [ ] `(2+x) - (1+x)` simplifies to `1`.
- [ ] `(a + b) - (a + b)` simplifies to `0`.
- [ ] Deterministic regression test added to `CoreBugFixTest` covering at least the three cases above (replaces the marker comment near `SubSymbolDispatchSmoke` in `tests/CoreBugFixTest.h`).
- [ ] Same fix applied symmetrically to the t2s and tensor specializations if they have parallel implementations.

## References

- Surfaced in PR #90, fifth-pass review and follow-up commit `bf75393`.
- Marker comment in `simplifier_sub.h` at the affected dispatch.
- Marker comment in `tests/CoreBugFixTest.h` near `SubSymbolDispatchSmoke`.
- Working reference for the coefficient-guard pattern: `merge_add` in `include/numsim_cas/functions.h`.

petlenz:
First pass on [#95](https://github.com/NumSim-Stack/numsim-cas/issues/95) — simplifier-layer audit and the directly-actionable findings. Closes #91.

## What this PR does

**Commit 1 — Audit.** Adds `docs/simplifier-coverage.md`. Inventories the generic dispatchers in `core/simplifier/`, the per-domain wrappers, and ranks the asymmetries by impact. Two material gaps were filed as follow-up issues during the audit (#96, #97). Four other findings are documented in the doc as either deliberate architectural choices or low-priority watch-list items.

**Commit 2 — Fix #91.** Two defects in `n_ary_sub_dispatch::dispatch(add, add)`:

1. `lhs.coeff() + rhs.coeff()` was unguarded against invalid coefficients and threw `invalid_expression_error` on the common case (e.g. `(x+y+z) - (x+y)`).
2. Children matching by key combined via `+`, and rhs-only children were pushed without negation. Both should be `-` for subtraction.

The corrected dispatch combines coefficients with `-` (guarded, mirroring the `merge_add` pattern in `functions.h`), combines matching children via `-` and filters zero results, negates rhs-only children, and collapses trivial results (empty children → coefficient; single child + no coeff → that child).

`SubMergeAddDispatchCorrect` in `CoreBugFixTest` covers the three #91 acceptance criteria:
- `(x+y+z) - (x+y) → z`
- `(2+x) - (1+x) → 1`
- `(a+b) - (a+b) → 0`

Confirmed to fail without the fix and pass with it.

## What this PR does **not** do

Per the audit, the bigger refactors spin off as separate issues rather than being attempted here:
- #96 — missing tensor pow simplifier.
- #97 — missing `n_ary_mul_mul_dispatch` (LHS=mul case for `*`).
- Five small duplication candidates between add and sub branches — listed in the audit as a watch list, not extracted (drift-risk vs API-clarity doesn't favor extracting 3–5 line blocks today).

## Base

Stacked on `remove-flaky-skip-list` (PR #90). Re-target to `main` once #90 lands.

## Test plan

- [x] `ctest` — 100% pass (1495 → 1496 with the new `SubMergeAddDispatchCorrect` regression)
- [x] `SubMergeAddDispatchCorrect` confirmed to fail before the fix and pass after
- [x] Audit doc cross-referenced with `grep` against actual source for each dispatcher class

### DIFF
--- a/docs/simplifier-coverage.md
+++ b/docs/simplifier-coverage.md
@@ -1,0 +1,175 @@
+# Simplifier Coverage
+
+> Audit of the simplifier layer across the three expression domains, captured
+> for issue [#95](https://github.com/NumSim-Stack/numsim-cas/issues/95). The
+> goal is to surface drift, missing rules, and architectural asymmetries so
+> they can be tracked rather than rediscovered.
+>
+> **Scope**: construction-time and visitor-driven simplification in
+> `include/numsim_cas/{core,scalar,tensor,tensor_to_scalar}/simplifier/`. Does
+> not cover differentiation, evaluation, printing, or substitution visitors.
+
+## Layering
+
+```
+core/simplifier/      ← generic dispatchers (templated on Traits)
+scalar/simplifier/    ← scalar wrappers (inherit core)
+tensor/simplifier/    ← tensor wrappers (hand-written; tensor mul_type is void)
+tensor_to_scalar/simplifier/  ← t2s wrappers (inherit core + unwrap pattern)
+```
+
+The generic dispatchers parameterise over `domain_traits<Domain>`
+([`include/numsim_cas/core/domain_traits.h`](../include/numsim_cas/core/domain_traits.h)),
+which exposes per-domain type aliases (`add_type`, `mul_type`, `negative_type`,
+…) and numeric hooks (`try_numeric`, `zero`, `one`, `make_constant`). Specialisations
+live in `scalar/scalar_domain_traits.h` and `tensor/tensor_domain_traits.h`.
+
+## Generic dispatcher inventory
+
+### Add — 7 dispatchers ([`simplifier_add.h`](../include/numsim_cas/core/simplifier/simplifier_add.h))
+
+| Dispatcher | LHS | Key overloads (rhs type) | Source |
+|---|---|---|---|
+| `add_dispatch` | base | `Expr` (fallback), `zero_type`, `negative_type`, `add_type` | `:19` |
+| `constant_add_dispatch` | constant | `constant_type`, `add_type`, `one_type`, `negative_type` | `:111` |
+| `one_add_dispatch` | one | `constant_type`, `add_type`, `one_type`, `negative_type` | `:185` |
+| `n_ary_add_dispatch` | add (n-ary sum) | `constant_type`, `one_type`, `symbol_type`, `add_type`, `negative_type` | `:245` |
+| `n_ary_mul_add_dispatch` | mul | `symbol_type`, `mul_type` | `:349` |
+| `symbol_add_dispatch` | symbol | `symbol_type`, `mul_type` | `:403` |
+| `negative_add_dispatch` | negative | `negative_type`, `add_type`, `constant_type` | `:452` |
+
+### Sub — 7 dispatchers ([`simplifier_sub.h`](../include/numsim_cas/core/simplifier/simplifier_sub.h))
+
+| Dispatcher | LHS | Key overloads (rhs type) | Source |
+|---|---|---|---|
+| `sub_dispatch` | base | `Expr` (fallback), `zero_type` | `:18` |
+| `negative_sub_dispatch` | negative | `add_type` | `:83` |
+| `constant_sub_dispatch` | constant | `constant_type`, `add_type`, `one_type` | `:114` |
+| `n_ary_sub_dispatch` | add | `constant_type`, `one_type`, `symbol_type`, `add_type` | `:169` |
+| `n_ary_mul_sub_dispatch` | mul | `symbol_type`, `mul_type` | `:268` |
+| `symbol_sub_dispatch` | symbol | `symbol_type`, `mul_type` | `:345` |
+| `one_sub_dispatch` | one | `constant_type` | `:396` |
+
+Naming is parallel to the add side, but **overload set is narrower** — e.g.
+`negative_sub_dispatch` only handles `add_type` rhs (vs add side's four).
+
+### Mul — 1 dispatcher ([`simplifier_mul.h`](../include/numsim_cas/core/simplifier/simplifier_mul.h))
+
+| Dispatcher | LHS | Key overloads (rhs type) | Source |
+|---|---|---|---|
+| `mul_dispatch` | base | `Expr` (fallback), `zero_type`, `one_type`, `mul_type` | `:16` |
+
+**No LHS-specialised siblings.** Every rule for multiplication of compound
+expressions (e.g. mul × mul, symbol × mul) goes through the base dispatcher's
+`get_default()` path or is handled per-domain. See *Asymmetry findings* below.
+
+### Pow — 3 dispatchers ([`simplifier_pow.h`](../include/numsim_cas/core/simplifier/simplifier_pow.h))
+
+| Dispatcher | LHS | Key overloads (rhs type) | Source |
+|---|---|---|---|
+| `pow_dispatch` | base | `Expr` (fallback) | `:15` |
+| `pow_pow_dispatch` | pow | `Expr`, `negative_type` | `:67` |
+| `mul_pow_dispatch` | mul | `negative_type`, `Expr` | `:97` |
+
+## Per-domain wrappers
+
+### Scalar — full coverage, 1 domain-specific rule
+
+| Family | Wrapper classes | Notes |
+|---|---|---|
+| Add | `add_default_visitor`, `constant_add`, `add_scalar_one`, `n_ary_add`, `n_ary_mul_add`, `symbol_add`, `add_negative`, **`pow_add`** | All inherit matching `*_dispatch<scalar_traits>`. `pow_add` adds `sin²(x) + cos²(x) → 1` ([`scalar_simplifier_add.h:127`](../include/numsim_cas/scalar/simplifier/scalar_simplifier_add.h)) — domain-specific. |
+| Sub | `sub_default_visitor`, `negative_sub`, `constant_sub`, `n_ary_sub`, `n_ary_mul_sub`, `symbol_sub`, `scalar_one_sub` | All inherit matching `*_dispatch<scalar_traits>`. No added overloads. |
+| Mul | `mul_default<Derived>`, `constant_mul`, `n_ary_mul`, `exp_mul`, `scalar_pow_mul`, `symbol_mul` | `exp_mul`: `exp(a)·exp(b) → exp(a+b)`. `scalar_pow_mul`: `pow(x,a)·pow(x,b) → pow(x,a+b)`. |
+| Pow | `pow_default<Derived>`, `pow_pow`, `mul_pow` | Inherits generic; subclasses specialise. |
+
+### Tensor — hand-written; no pow simplifier
+
+| Family | Wrapper classes | Notes |
+|---|---|---|
+| Add | `add_default<Derived>`, `n_ary_add`, `symbol_add`, **`tensor_scalar_mul_add`**, `add_negative` | **Does not inherit generic.** Hand-written because `tensor_traits::mul_type` is `void` (no n-ary scalar multiplication node in this domain — see [`tensor_domain_traits.h`](../include/numsim_cas/tensor/tensor_domain_traits.h)). `tensor_scalar_mul_add` is domain-specific. |
+| Sub | `sub_default<Derived>`, `negative_sub`, `n_ary_sub`, `symbol_sub` | Same architectural note as add. |
+| Mul | `mul_default<Derived>`, `tensor_pow_mul`, `kronecker_delta_mul`, `symbol_mul`, `n_ary_mul` | Hand-written. `kronecker_delta_mul` handles index contraction; `symbol_mul`: `x·x → pow(x,2)`. |
+| Pow | **— absent —** | No `tensor_simplifier_pow.h` file. Rules like `pow(pow(A,m), n) → pow(A, m·n)` and `pow(inv(A), n)` have no construction-time handling. → tracked separately. |
+
+Additional domain-specific simplifiers (not in the dispatcher hierarchy):
+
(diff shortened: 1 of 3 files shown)


petlenz:
Self-review of PR #98. Most of this is solid; a few items worth surfacing.

## Real concerns

### Audit doc scope statement is wrong

`docs/simplifier-coverage.md` opens with:

> Scope: construction-time and visitor-driven simplification in `include/numsim_cas/{core,scalar,tensor,tensor_to_scalar}/simplifier/`.

The doc actually only inventories the **visitor-driven** dispatchers. Construction-time simplifications live elsewhere and are not in the matrix:

- `tensor_functions.h::inv()` (inv-of-inv, inv-of-I, inv-of-kronecker, the new skew rejection)
- `tensor_functions.h::trans()` (trans-of-trans)
- `scalar_std.h::sin/cos/tan/asin/acos/atan/exp/log/pow/sqrt` (inverse pairs, constant folding, zero/one short-circuits)
- `tensor_to_scalar_functions.cpp::trace/det/norm/dot` (trace(0), det(I), etc.)
- The `tensor_operators.h` add/sub `is_trans_of` annotations from PR #90.

Either expand the audit to cover those (significant work; probably out of scope here) or narrow the scope statement to "visitor-driven simplification in `*/simplifier/`". Recommend the latter for this branch, with a note that construction-time simplifications deserve a parallel audit (potential follow-up issue).

### `dispatch(add_type)` return-type contract is broader than the comment admits

Previously the dispatch returned the freshly-built `Traits::add_type` unconditionally. The new version can return:

- `Traits::add_type` — normal case
- `Traits::zero_type` — coeff and all children cancelled
- The coefficient expression — children cancelled, coeff survived
- A child expression directly — one child survived, no coeff

This is correct (matches what callers expect via the visitor pattern, which just propagates the holder), but the doc comment only mentions "the coefficient (or zero) is returned directly." Worth spelling out the four cases since a future reader instrumenting the dispatch may assume `add_type` and trip on the others.

### Trivial-result collapse is bespoke

The collapse block at the bottom of the dispatch:

```cpp
if (add.symbol_map().empty()) {
  return add.coeff().is_valid() ? add.coeff() : Traits::zero();
}
if (add.symbol_map().size() == 1 && !add.coeff().is_valid()) {
  return add.symbol_map().begin()->second;
}
return expr;
```

…is local to this dispatch but the same pattern applies to every place that builds an `n_ary_tree`-backed add and might end up trivial. `n_ary_add_dispatch::dispatch(symbol_type)` and the analogous merge paths could produce empty-children adds too (rarer in practice since `+` doesn't naturally cancel children) but the absence of collapse means they'd return a malformed `add_type` carrier instead of the bare expression.

If this pattern is going to keep recurring, it belongs as a member on `n_ary_tree` (e.g. `finalize_or_collapse()` returning `expr_holder_t`) or as a free helper in `simplifier_common.h`. Not a blocker — the dispatch fix is correct as-is — but flag for the broader unification work (#95).

### Zero-detection assumes the canonical zero singleton

`is_same<typename Traits::zero_type>(...)` returns true only when the result has type-id `scalar_zero` / `tensor_zero` / `tensor_to_scalar_zero`. The check works for symbol-vs-symbol cancellation (e.g. `x - x → scalar_zero`) and equal-constant cancellation, because the relevant simplifiers short-circuit to the canonical zero singleton.

It does **not** catch mathematically-zero results that simplification doesn't reduce to the singleton. For example, if a non-trivial expression `E - E` somehow took a path that produced `make_constant(0)` rather than `scalar_zero`, my filter would let it through and leave a constant-zero child in the result. The current simplifier pipeline appears to always normalise to the singleton, so this is hypothetical — but a comment noting the assumption would prevent surprise if normalisation rules ever drift.

### No tensor-to-scalar regression test

The fix is in a template instantiated by both `scalar_traits` and `tensor_to_scalar_traits` (tensor's `mul_type` is void, so it's not instantiated there — confirmed by the audit). The new `SubMergeAddDispatchCorrect` test covers only the scalar path. The t2s path *should* work identically because the generic dispatcher is the same code, but no test locks that in. Worth adding a t2s analog: build a `dot(A,A) + trace(B)` style add, subtract a partially-overlapping t2s add, assert the cancellation behaves.

## Smaller

### Audit doc — watch-list item 5 is now misleading

Pair 5 in the watch list (`simplifier_add.h:291` vs `simplifier_sub.h:243`) — the add-side and sub-side `dispatch(symbol_type)` paths — was a true mirror before this PR. Commit `685afa3` doesn't touch `dispatch(symbol_type)`, so the pair is still valid, but the new `dispatch(add_type)` in `simplifier_sub.h:217` has now significantly diverged from its add-side mirror (`simplifier_add.h:306`) — it has zero-filtering, trivial-result collapse, and validity-guarded coeff handling that the add side legitimately doesn't need. Update the doc to note this divergence is intentional (sub naturally produces zero children; add doesn't).

### Audit doc — t2s "scalar-domain op" claim is unverified

The t2s row says "Domain-specific: `n_ary_add::dispatch(tensor_to_scalar_scalar_wrapper)` unwraps the wrapper, runs the scalar-domain op, rewraps." I copied that wording from the exploration agent's output without verifying. Worth a 30-second look at `tensor_to_scalar_simplifier_add.cpp` to confirm whether it runs scalar-domain `+` (likely) or t2s-domain `+`, and update the doc accordingly.

### Test naming

`SubMergeAddDispatchCorrect` doesn't grep nicely against the dispatcher name. `NArySubAddDispatchCancelsCorrectly` or `SubTwoAdds` would be more discoverable.

## Confirmed good

- Audit captures the real asymmetries; #96 and #97 are well-scoped follow-ups.
- The #91 fix passes the three acceptance cases and was verified to fail without the fix.
- Marker comments from PR #90 are cleaned up alongside the fix.
- No behavioural changes outside the targeted dispatch.

## Recommendation

Address the audit-doc scope statement (the misleading "construction-time" claim) and add a t2s regression. The other points are worth noting in commit messages or follow-up issues but don't block the PR.

petlenz:
Pushed `871380d` addressing the self-review.

- **Audit doc scope statement** narrowed to "visitor-driven simplification in `*/simplifier/` only." Construction-time simplifications (`inv(inv(A))→A`, `sin(asin(x))→x`, `trace(0)→0`, the `is_trans_of` annotations) are explicitly out of scope with a note that they deserve a parallel audit. Watch-list entry for the merge-two-adds pair updated to mark the post-PR #98 divergence (zero-filtering + collapse) as legitimate, not drift.
- **Return-type contract** spelled out in the `dispatch(add_type)` comment — four possible return types, with a note that callers via the visitor pattern propagate the holder transparently.
- **Canonical-zero-singleton assumption** documented in the same comment block. Filter is reliable today; comment flags the assumption so a future reader knows what could drift.
- **T2s regression added.** `NArySubAddDispatchT2s` covers `(trace(X)+trace(Y)+trace(Z)) - (trace(X)+trace(Y)) → trace(Z)` and the `→ 0` case. The original test is renamed `NArySubAddDispatchScalar` to grep against the dispatcher class name.
- **Centralisation follow-up filed as #99** — the trivial-result collapse pattern belongs in `n_ary_tree` (or `simplifier_common.h`) so other dispatchers can adopt it as needed. Cross-referenced from the audit doc.

T2s "scalar-domain op" claim in the doc verified against `tensor_to_scalar_simplifier_add.cpp:69-94` (`existing_w.expr() + rhs_w.expr()` is the scalar-domain `+`); doc wording stands.

1497 tests pass.

petlenz:
Second pass.

## Confirmed addressed

- Audit doc scope narrowed correctly; the construction-time note is in place and the watch-list pair update accurately marks the new divergence as legitimate. ✓
- Dispatch doc comment now lists all four return-type cases and documents the canonical-zero assumption. ✓
- T2s regression `NArySubAddDispatchT2s` exercises the t2s template instantiation cleanly. The two cases (single survivor → returns child via the size-1 collapse; full cancellation → returns `tensor_to_scalar_zero` via the empty collapse) hit different paths in the same dispatch. ✓
- Test names are now grep-friendly against `n_ary_sub_dispatch`. ✓
- Centralisation follow-up filed as #99. ✓
- T2s "scalar-domain op" claim was actually verified against `tensor_to_scalar_simplifier_add.cpp:69-94` (`existing_w.expr() + rhs_w.expr()` uses scalar `+`); the doc stays. ✓

## One small thing worth noting

The dispatch could in principle still be reached via an LHS=add path with a rhs that's an `add` whose `mul_type` is reachable as a child. In the t2s domain, `dot(X, Y)` returns a `tensor_to_scalar_mul` (RHS of an outer add); subtracting two such adds where the children are themselves `tensor_to_scalar_mul`s exercises an indirect path. The test only uses `trace(X)` (which produces a `tensor_to_scalar_trace` symbol-like node, not a `_mul`), so the t2s test doesn't cover the case where the merge-via-`-` operates on `_mul` children that themselves carry coefficients.

Probably not worth adding here — the path goes through the same `merge_or_insert(child - pos->second)` logic and `child - pos->second` delegates to the t2s sub operator, which has its own dispatch. The test as-is locks in the dispatch-level contract. Worth a one-liner mention if you want to be thorough.

## Honest assessment

PR is ready. The simplifier-coverage doc, the #91 fix, and the regressions all hang together. The remaining items from my prior reviews are either tracked (#92, #93, #94, #96, #97, #99) or out-of-scope additions I'd flag with no good reason to require.

Approving.

petlenz:

After four rounds: audit doc, #91 fix, the `is_trans_of` shared helper, t2s regression. 1497 tests pass at tip. The downstream chain (#100, #103, #105, #106, #107, #108) is reviewed and stacked.

No new findings. Ready to merge from my side.

PR #100, opened 2026-05-15
Centralize trivial-result collapse as finalize_add<Traits> helper

### LINKED ISSUE #99: Centralize trivial-result collapse pattern from simplifier dispatchers
## Summary

PR #98 added a "trivial-result collapse" block to `n_ary_sub_dispatch::dispatch(add, add)` in `include/numsim_cas/core/simplifier/simplifier_sub.h`:

```cpp
if (add.symbol_map().empty()) {
  return add.coeff().is_valid() ? add.coeff() : Traits::zero();
}
if (add.symbol_map().size() == 1 && !add.coeff().is_valid()) {
  return add.symbol_map().begin()->second;
}
return expr;
```

The pattern applies to any dispatcher that constructs an `n_ary_tree`-backed expression and could end up with a trivial result (empty children, or a single child with no coefficient). Today only the sub-side merge-two-adds path explicitly produces these trivial results often enough to need the collapse — `+` rarely cancels its children — but as more simplification rules are added (especially around projector algebra and the missing tensor pow simplifier in #96), other dispatchers will need the same logic.

Surfaced during the PR #98 review for issue #95.

## Proposal

Centralise the collapse as either:

1. **A member on `n_ary_tree`**: `expression_holder<expr_t> finalize_or_collapse() &&` (rvalue-qualified — collapse-on-finalisation). The dispatcher writes `return add.finalize_or_collapse()` at the end. This couples the helper to the tree type and is the most direct fit.
2. **A free helper** in a new `include/numsim_cas/core/simplifier/simplifier_common.h` taking the tree by ref + a `Traits::zero_type` hook. More flexible but more verbose at the call site.

Option 1 reads better at the call site; option 2 makes the Traits dependency explicit. Either works.

## Acceptance criteria

- [ ] Move the collapse logic from `simplifier_sub.h::n_ary_sub_dispatch::dispatch(add_type)` to the chosen helper location.
- [ ] Audit other dispatchers in `core/simplifier/` for places that construct an n_ary_tree and might benefit from the same collapse. Apply where appropriate.
- [ ] Add a unit test for the helper itself (covering: empty-children-with-coeff → coeff; empty-children-without-coeff → zero; one-child-no-coeff → child; non-trivial → unchanged).
- [ ] Verify behaviour preserved by `tests/CoreBugFixTest.h::NArySubAddDispatch*`.

## References

- `include/numsim_cas/core/simplifier/simplifier_sub.h` — current location of the inlined pattern (the `dispatch(add_type)` body).
- `include/numsim_cas/core/n_ary_tree.h` — where the helper would naturally live for option 1.
- PR #98 — context for why this surfaced now.
- Issue #95 — simplifier unification umbrella.
- Issue #96 — tensor pow simplifier (likely consumer of the same pattern).
- Issue #97 — `n_ary_mul_mul_dispatch` (likely consumer).

petlenz:
Closes #99. Stacked on #98.

Extracts the trivial-result collapse logic that PR #98 inlined into `n_ary_sub_dispatch::dispatch(add_type)` into a free helper template:

```cpp
// include/numsim_cas/core/simplifier/simplifier_common.h
template <typename Traits>
[[nodiscard]] expression_holder<typename Traits::expression_type>
finalize_add(expression_holder<typename Traits::expression_type> expr);
```

Handles four shapes:

| Shape | Returns |
|---|---|
| empty children + valid coeff | the coefficient |
| empty children + invalid coeff | `Traits::zero()` |
| one child + invalid coeff | the child |
| otherwise | the freshly-built add unchanged |

Free function rather than member on `n_ary_tree` because the zero fallback is domain-specific (`Traits::zero()`), which `n_ary_tree` has no access to.

## Applied to

- `n_ary_sub_dispatch::dispatch(add_type)` — the original site (replaced the inlined block from PR #98).
- `n_ary_sub_dispatch::dispatch(SymbolType)` — `(x+y+z) - x` shape. Consistent with the other dispatch; `(x) - x` can now collapse to zero properly via the helper.

## Tests

Five direct unit tests on the helper itself (`tests/CoreBugFixTest.h`):
- empty + coeff → coeff
- empty + no coeff → zero
- one child + no coeff → child
- non-trivial → unchanged
- one child + coeff (e.g. `1+x`) → unchanged (not a trivial shape)

Existing integration tests (`NArySubAddDispatchScalar`, `NArySubAddDispatchT2s`) still pass.

## Test plan

- [x] `ctest` — 100% pass (1497 → 1502 with five new helper tests)

## Base

### DIFF
--- a/include/numsim_cas/core/simplifier/simplifier_common.h
+++ b/include/numsim_cas/core/simplifier/simplifier_common.h
@@ -1,0 +1,44 @@
+#ifndef NUMSIM_CAS_CORE_SIMPLIFIER_COMMON_H
+#define NUMSIM_CAS_CORE_SIMPLIFIER_COMMON_H
+
+#include <numsim_cas/core/expression_holder.h>
+
+namespace numsim::cas::detail {
+
+// finalize_add: collapses an `n_ary_tree`-backed add expression to a simpler
+// form when its content is trivial. Used at the end of dispatchers that
+// build an add from scratch (e.g. n_ary_sub_dispatch::dispatch(add_type),
+// where children matching by key can fully cancel).
+//
+// Returns one of:
+//   - the coefficient expression — if the symbol_map is empty and the
+//     coefficient is valid;
+//   - Traits::zero() — if the symbol_map is empty AND the coefficient is
+//     invalid (i.e. nothing remained after cancellation);
+//   - the single child — if the symbol_map has exactly one child and no
+//     coefficient is set;
+//   - the original `expr` unchanged otherwise.
+//
+// Callers via the visitor pattern propagate the holder transparently, so
+// the return-type broadening is invisible to them.
+//
+// Why a free helper rather than a member on n_ary_tree: the zero-fallback
+// is domain-specific (Traits::zero()), which n_ary_tree doesn't have access
+// to. This helper sits in the simplifier layer where Traits is already in
+// scope at the call site.
+template <typename Traits>
+[[nodiscard]] inline expression_holder<typename Traits::expression_type>
+finalize_add(expression_holder<typename Traits::expression_type> expr) {
+  auto &add = expr.template get<typename Traits::add_type>();
+  if (add.symbol_map().empty()) {
+    return add.coeff().is_valid() ? add.coeff() : Traits::zero();
+  }
+  if (add.symbol_map().size() == 1 && !add.coeff().is_valid()) {
+    return add.symbol_map().begin()->second;
+  }
+  return expr;
+}
+
+} // namespace numsim::cas::detail
+
+#endif // NUMSIM_CAS_CORE_SIMPLIFIER_COMMON_H
--- a/include/numsim_cas/core/simplifier/simplifier_common.h
+++ b/include/numsim_cas/core/simplifier/simplifier_common.h
@@ -27,7 +27,7 @@ namespace numsim::cas::detail {
 // to. This helper sits in the simplifier layer where Traits is already in
 // scope at the call site.
 template <typename Traits>
-[[nodiscard]] inline expression_holder<typename Traits::expression_type>
+[[nodiscard]] expression_holder<typename Traits::expression_type>
 finalize_add(expression_holder<typename Traits::expression_type> expr) {
   auto &add = expr.template get<typename Traits::add_type>();
   if (add.symbol_map().empty()) {
--- a/include/numsim_cas/core/simplifier/simplifier_sub.h
+++ b/include/numsim_cas/core/simplifier/simplifier_sub.h
@@ -4,6 +4,7 @@
 #include <numsim_cas/basic_functions.h>
 #include <numsim_cas/core/domain_traits.h>
 #include <numsim_cas/core/scalar_number.h>
+#include <numsim_cas/core/simplifier/simplifier_common.h>
 #include <ranges>
 #include <set>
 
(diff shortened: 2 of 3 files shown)


petlenz:
Self-review of PR #100 / issue #99.

## Worth noting

**`dispatch(SymbolType)` latent zero-leak surfaced and fixed in commit `c803f8b`.** During the audit-mandated "apply the helper where appropriate" step, I discovered the symbol-dispatch path had the same zero-leak shape as the original `dispatch(add_type)` bug (#91) — `(2+x)-x` produced `"2 + 0"` instead of `"2"`. Adding the helper exposed it (finalize_add doesn't collapse multi-child results, even when one of those children is a stray zero). Fixed with the same one-line zero-filter pattern as `dispatch(add_type)`. Direct regression locked in by `NArySubSymbolDispatchCancelsCleanly`.

The fact that this latent bug survived all the way to PR #100 is itself a small finding — the existing `SubSymbolDispatchSmoke` test in CoreBugFixTest only verified no-throw, not result correctness. Worth a follow-up: audit all `*DispatchSmoke` tests for "asserts only that construction doesn't throw" and upgrade them where the actual value matters.

## Smaller

**`NArySubSymbolDispatchCancelsCleanly` third case `(x+x) - x - x → 0` may route through a different dispatcher.** `(x+x)` simplifies to `2*x` (a `scalar_mul`), so the subsequent `- x` hits `n_ary_mul_sub_dispatch::dispatch(symbol_type)`, not the path this test is named for. The assertion is still meaningful (result correctness) but the path-coverage claim in the test name is misleading. Either rename or replace with `(x+y) - x → y` which definitely exercises the targeted path. Cosmetic.

**Documentation duplication.** The "canonical zero singleton" caveat from `dispatch(add_type)`'s doc comment is not repeated at the new `dispatch(SymbolType)` zero-filter. Could move the caveat to `simplifier_common.h::finalize_add` or to a shared internal note. Low priority.

**`inline` is redundant on a function template.** `finalize_add` declares both `[[nodiscard]] inline`. Function templates have implicit inline linkage. Harmless but stylistically noisy.

## Confirmed good

- Helper is correctly scoped to `numsim::cas::detail` matching siblings.
- Free template avoids coupling `n_ary_tree` to `Traits::zero()`. Rationale documented in the helper.
- Five direct unit tests cover the four trivial shapes + the "one child + valid coeff" non-trivial case.
- `NArySubAddDispatchScalar` / `NArySubAddDispatchT2s` (the PR #98 regressions) still pass — substitution preserves behavior on those inputs.

## No follow-ups required

The audit-mandated "apply where appropriate" is done for the two `simplifier_sub.h` sites. Other simplifier dispatchers (`simplifier_add.h`, etc.) don't build trivial-result adds in the same way (`+` doesn't naturally cancel children to zero), so the helper isn't applicable there.

petlenz:
Second-pass review.

## Addressed since the first review

- Dropped redundant `inline` on `finalize_add` template. ✓
- Replaced the misleading `(x+x)-x-x` third test case (which routes through `n_ary_mul_sub_dispatch`, not the target) with `(x+y) - x → y` which definitely exercises `n_ary_sub_dispatch::dispatch(SymbolType)`. ✓
- New finding from this pass: `dispatch(SymbolType)` not-found branch used `push_back(-m_rhs)` which threw on duplicate-child collision (e.g. `(-x + y) - x`). Switched to `merge_or_insert(-m_rhs)`. Regression `NArySubSymbolDispatchNotFoundCombinesWithExisting` locks it in. ✓

## Found and filed as separate issues

- **#101** — audit `*DispatchSmoke` / `*ConstructionSmoke` tests for missing value assertions. The latent zero-leak this PR fixed survived because `SubSymbolDispatchSmoke` only checked no-throw.
- **#102** — `constant_sub_dispatch::dispatch(add_type)` has the same unguarded-coeff + push_back bug pattern. Probed with `2 - (x + y)`: throws. Same shape as #91 / PR #98. Out of scope for #99 (different class); filed separately.

## Honest assessment

PR is consistent now. All three review-loop findings are either fixed in-PR (when in the same dispatcher I was already touching) or filed as separate issues (when in a different dispatcher). No new issues with the centralisation itself.

Approving.

petlenz:

After two rounds: extracted `finalize_add<Traits>`, found and fixed two latent bugs in `dispatch(SymbolType)` (zero leak + push_back collision). 1507 tests pass.

No new findings on this pass. Ready to merge.

PR #103, opened 2026-05-15
Correct #97 audit claim; lock in existing mul × mul merge behavior

### LINKED ISSUE #97: Add n_ary_mul_mul_dispatch (LHS=mul) sibling for the mul dispatcher family

petlenz:
Closes #97.

Investigation during the #97 work showed the audit's claim that `(c₁·x·y) · (c₂·x·z) → c₁·c₂·pow(x,2)·y·z` "isn't reachable without a dedicated `n_ary_mul_mul_dispatch`" was **wrong**. The rules ARE implemented today, just in per-domain wrappers rather than a generic dispatcher.

## What's actually there

| Domain | Where | Approach |
|---|---|---|
| Scalar | `n_ary_mul::dispatch(scalar_mul)` in `scalar_simplifier_mul.cpp:100` | Operator-chain: copy lhs, multiply each rhs child via `operator*` (re-triggers simplification) |
| T2s | `mul_base::dispatch(tensor_to_scalar_mul)` in `tensor_to_scalar_simplifier_mul.cpp:99` | `push_or_combine` helper: inlined merge with pow-promotion |
| Tensor | n/a | Different mul semantics (inner-product, non-commutative) |

Verified by direct probe:
- `(2*x*y) * (3*x*z) = 6*pow(x,2)*y*z` ✓
- `(x*y) * (x*z) = pow(x,2)*y*z` ✓
- `(x*y*z) * (x*a) = a*pow(x,2)*y*z` ✓
- `(tr(X)*tr(Y)) * (tr(X)*tr(Z)) = pow(tr(X),2)*tr(Y)*tr(Z)` ✓

## What this PR does

1. **Updates the audit doc** (`docs/simplifier-coverage.md`) to correct the claim and document where the mul × mul logic actually lives in each domain.
2. **Adds two regression tests** in `CoreBugFixTest.h`:
   - `MulMulMergesLikeFactorsScalar`
   - `MulMulMergesLikeFactorsT2s`
   These lock in the contract so any future refactor that lifts the merge into a generic dispatcher can't silently drop it.

## What it deliberately does NOT do

The architectural concern that #95 raised is real: the **drift** between scalar's operator-chain approach and t2s's `push_or_combine` approach is a refactoring opportunity. But:
- Both approaches work correctly today.
- The work to consolidate is substantial (lift to a generic dispatcher, choose between the two patterns, migrate the per-domain wrappers).
- The benefit is marginal — minor perf gain for scalar, architectural symmetry.

Worth filing as a separate refactor issue if pursued; not a missing-feature gap as #97 claimed.

## Base

## Test plan

- [x] `ctest` — 100% pass (1497 → 1499 with the two new regressions)

### DIFF
--- a/docs/simplifier-coverage.md
+++ b/docs/simplifier-coverage.md
@@ -125,12 +125,33 @@ a tracking issue.
    (`pow(pow(A, m), n) → pow(A, m·n)`, `pow(inv(A), n) → inv(pow(A, n))`, etc.).
    Tracked as [#96](https://github.com/NumSim-Stack/numsim-cas/issues/96).
 
-2. **No `n_ary_mul_mul_dispatch`.** The mul family has only the base
-   `mul_dispatch` ([`simplifier_mul.h:16`](../include/numsim_cas/core/simplifier/simplifier_mul.h)).
-   Rules like `(c₁·x·y) · (c₂·x·z) → c₁·c₂·pow(x,2)·y·z` aren't reachable
-   without a dedicated LHS=mul dispatcher. Compare with the add side's
-   `n_ary_add_dispatch`, which exists.
-   Tracked as [#97](https://github.com/NumSim-Stack/numsim-cas/issues/97).
+2. **No `n_ary_mul_mul_dispatch`** in the generic layer, but the
+   functional rules **are** implemented per-domain. The mul family in
+   `core/simplifier/simplifier_mul.h` has only the base `mul_dispatch`,
+   yet:
+   - **Scalar** handles mul × mul via `n_ary_mul::dispatch(scalar_mul)` in
+     [`scalar_simplifier_mul.cpp:100`](../src/numsim_cas/scalar/simplifier/scalar_simplifier_mul.cpp).
+     Approach: copy lhs, multiply each rhs child into the result via
+     `operator*` (re-triggers simplification).
+   - **T2s** handles mul × mul via `mul_base::dispatch(tensor_to_scalar_mul)`
+     in [`tensor_to_scalar_simplifier_mul.cpp:99`](../src/numsim_cas/tensor_to_scalar/simplifier/tensor_to_scalar_simplifier_mul.cpp).
+     Approach: copy lhs, call a `push_or_combine` helper for each rhs
+     child (inlined merge with pow-promotion).
+   - **Tensor** mul has different semantics (inner-product based, not
+     commutative) and doesn't apply.
+   
+   `(c₁·x·y) · (c₂·x·z) → c₁·c₂·pow(x,2)·y·z` works in both scalar and
+   t2s today (verified by `tests/CoreBugFixTest.h` regression tests
+   `MulMulMergesLikeFactorsScalar` / `T2s`).
+   
+   The architectural concern that #95 raised is real: the **drift**
+   between the scalar and t2s implementations (different approaches to
+   the same algebraic task) is the kind of thing that surfaces bugs
+   like #91 / #102. Consolidating into a single generic dispatcher is
+   worth tracking as a refactor but isn't a missing-feature gap.
+   Tracked as [#97](https://github.com/NumSim-Stack/numsim-cas/issues/97)
+   (closed as based on wrong premise) and a new follow-up if
+   consolidation is pursued.
 
 3. **`pow_add` Pythagorean identity is scalar-only.** `sin²(x) + cos²(x) → 1`
    is implemented in [`scalar_simplifier_add.h:127`](../include/numsim_cas/scalar/simplifier/scalar_simplifier_add.h)
--- a/tests/CoreBugFixTest.h
+++ b/tests/CoreBugFixTest.h
@@ -321,6 +321,11 @@ TEST(CoreBugFix, NArySubAddDispatchT2s) {
 // ---------------------------------------------------------------------------
 
 TEST(CoreBugFix, MulMulMergesLikeFactorsScalar) {
+  // EXPECT_EQ compares via expression_holder::operator==, which is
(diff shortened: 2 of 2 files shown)


petlenz:
Self-review of PR #103 / issue #97.

## Honest assessment

The PR's core contribution is correcting an audit-doc claim that was wrong on investigation, and adding two regression tests that lock in the existing (correct) behavior. That's a legitimate outcome of working off the issue — the investigation IS the work.

## Worth noting

**`a*pow(x,2)*y*z` ordering dependency.** `MulMulMergesLikeFactorsScalar` asserts `(x*y*z) * (x*a) == a*pow(x,2)*y*z`. The expected RHS uses alphabetical-looking ordering, which matches hash-based ordering on this platform. `operator==` on the holder uses hash equality, so the assertion is robust against ordering differences — but the *literal* form in the test code is platform-suggestive. If a reader copies the test for another case, they may build the wrong expected expression. Worth a one-line comment in the test noting that the order is enforced by hash, not by the literal.

**The "wrong premise" narrative.** I filed #97 based on my reading of the audit. The audit, in turn, was based on the exploration agent's report ("the mul family has only the base `mul_dispatch`"). The agent's observation was technically accurate (the **generic** layer has no LHS-specialised mul dispatcher) but the **functional** claim ("rules aren't reachable") was a leap. This is a real lesson about layering the audit findings: differentiate "architectural gap" from "functional gap." The audit doc now does this.

**Refactor opportunity acknowledged but not pursued.** The scalar and t2s implementations are doing the same algebraic task in different ways (operator-chain vs `push_or_combine`). The architectural symmetry concern from #95 is real. I deliberately didn't fix it because:
- Both work correctly.
- Choosing between approaches is a design decision deserving its own discussion.
- The scope of "consolidate two working implementations" is much larger than #97's stated scope.

A separate refactor issue would make sense if someone wants to pursue it.

## Confirmed good

- Audit doc correction is factually accurate (verified by reading the .cpp files I cite).
- Two regression tests pass and cover the contract.
- The PR closes #97 with a clear narrative — investigation found the claim was wrong, doc corrected, regressions added.

## No follow-ups required

The audit-mandated "apply the helper where appropriate" doesn't apply here since there's no helper to apply — the work was investigation + documentation correction. If anything, a refactor follow-up issue could track "unify scalar's operator-chain mul-mul merge with t2s's `push_or_combine` approach into a generic dispatcher." Whether that's worth doing is a judgment call I'm leaving to the maintainer.

Approving.

petlenz:

After two rounds: investigation found #97's premise was wrong, audit doc corrected, two regression tests lock in the existing mul × mul behavior. Filed #104 as the consolidation follow-up.

No new findings. Ready to merge.

Issue #93, opened 2026-05-15
Investigate platform-dependent Skew space annotation loss inside tensor_mul


petlenz:
## Summary

Commit `7e962e5` on the `remove-flaky-skip-list` branch (now merged via PR #90) added structural skew detection to the fuzz test with this rationale:

> The Skew space on `skew(G)` is not preserved on all platforms when stored inside `tensor_mul`. Detect skew structurally by checking if the expression is `inner_product(P_skew, ...)` where the LHS is a `tensor_projector` with `Skew` permutation.

PR #90 moved the structural classifier into the library (`include/numsim_cas/tensor/skew_classification.h`) and now uses it in `inv()`'s rejection. The structural path is the *authoritative* check; the space-annotation fast path is a hint.

But the **root cause of the platform-dependent annotation loss** was never identified, only worked around. As long as the annotation is unreliable, every other consumer that reads `data()->space()` is suspect — not just `inv()`.

## What we know

- `tensor_scalar_mul` and `tensor_negative` constructors propagate the `space()` from their inner expression (verified during the PR #90 review).
- `tensor_add::sub` operator annotates `trans(A) − A` / `A − trans(A)` and `trans(A) + (−A)` / `(−A) + trans(A)` as Skew at construction (added in PR #90).
- `tensor_mul` does **not** carry a Skew annotation derived from its children — it isn't generally true that a `*` of skew is skew, so that's correct.
- The platform dependence kicks in when a skew-annotated child is *stored inside* a `tensor_mul`. Per the commit message, the annotation on that child is sometimes not preserved on retrieval on some platforms.

## Hypotheses worth investigating

1. **`symbol_map` iteration order across stdlib implementations.** If any rewrite path reconstructs an annotated expression by iterating its children, libc++ vs libstdc++ could surface a different "first" child whose annotation gets preserved while others get dropped.
2. **Annotation lost during a rebuild path** (`tensor_rebuild_visitor` or similar). If the rebuild copies structure but doesn't propagate `space()`, then anything that goes through it loses the annotation.
3. **`make_expression<tensor_mul>` constructor** doesn't read children's `space()`, and a child whose annotation was set via `set_space()` *after* its hash was computed could lose it across some intermediate operation (related to the post-construction mutation we documented in PR #90).
4. **Specific simplifier paths** in `tensor_simplifier_mul.cpp` / `tensor_simplifier_add.cpp` that recurse via `accept` and rebuild — they might not be preserving the annotation on the result.

[...]

PR #108, opened 2026-05-15
Investigate #93: lock in Skew preservation (cannot reproduce original loss)

petlenz:
Partial work on #93.

## Honest framing

The platform-dependent Skew annotation loss that commit `7e962e5` flagged **cannot be reproduced on this build** (Linux / libstdc++ / gcc). Every composition path tested preserves the Skew annotation. The structural classifier added in PR #90 remains as the defensive fallback for the platforms where annotation propagation does break.

## Investigation

Probed these composition paths:

| Construction | Annotation preserved? |
|---|---|
| `skew(A)` | ✓ (constructor sets Skew) |
| `-skew(A)` | ✓ (tensor_negative preserves) |
| `2 * skew(A)` | ✓ (tensor_scalar_mul preserves) |
| `trans(skew(A))` | ✓ (resolves to `-skew(A)`, Skew preserved) |
| `skew(A) * B` as a `tensor_mul` child | ✓ (child retains Skew) |

If the loss exists, it's specific to a platform (libc++ vs libstdc++? specific gcc version? CI runner?) I don't have access to. Without a reproducer, I can't trace and fix at the source.

## What this PR does

Three lock-in tests in `CoreBugFixTest`:
- `SkewSpacePreservedThroughNegation`
- `SkewSpacePreservedThroughScalarMul`
- `SkewSpacePreservedAsTensorMulChild`

These narrow the surface where a future annotation-propagation regression could go undetected. They pass today on this build; they would fail on a build where the platform-specific loss recurs.

## What this PR does NOT do

- **Fix the original platform-specific loss.** Without reproduction, no source-level fix is possible.
- **Remove the structural fallback.** `skew_classification.h` stays as the authoritative check — the issue body's "re-evaluate whether `is_provably_skew` still needs the structural fallback" depends on the source-level fix, which is deferred.

## What's worth doing as follow-up

If the platform-specific loss is observed again on CI or another platform:
1. Capture the build environment (compiler, stdlib, flags).
2. Run the three lock-in tests on that environment — they should fail and isolate the path.
3. Trace from there.

The issue stays open as the tracker for that future investigation.

## Base

## Test plan

- [x] `ctest` — 100% pass (1497 → 1500 with the three new tests)

### DIFF
--- a/tests/CoreBugFixTest.h
+++ b/tests/CoreBugFixTest.h
@@ -315,6 +315,68 @@ TEST(CoreBugFix, NArySubAddDispatchT2s) {
 }
 
 // ---------------------------------------------------------------------------
+// Skew annotation propagation lock-in (issue #93).
+// On the build platform that motivated commit 7e962e5, the Skew space
+// annotation could be lost when skew(A) was stored inside a tensor_mul.
+// The structural skew classifier in skew_classification.h was added as
+// a defensive fallback. These tests lock in that on THIS build the
+// annotation IS preserved through the common composition paths — any
+// regression would be visible here (whether or not the platform-specific
+// loss the original commit observed ever recurs).
+// ---------------------------------------------------------------------------
+
+namespace {
+template <typename Holder> bool is_skew_annotated(Holder const &e) {
+  if (auto const &sp = e.get().space())
+    return std::holds_alternative<Skew>(sp->perm);
+  return false;
+}
+} // namespace
+
+TEST(CoreBugFix, SkewSpacePreservedThroughNegation) {
+  auto [A] =
+      make_tensor_variable(std::tuple{"A", std::size_t{3}, std::size_t{2}});
+  auto sA = skew(A);
+  ASSERT_TRUE(is_skew_annotated(sA));
+  EXPECT_TRUE(is_skew_annotated(-sA));
+}
+
+TEST(CoreBugFix, SkewSpacePreservedThroughScalarMul) {
+  auto [A] =
+      make_tensor_variable(std::tuple{"A", std::size_t{3}, std::size_t{2}});
+  auto sA = skew(A);
+  ASSERT_TRUE(is_skew_annotated(sA));
+  EXPECT_TRUE(is_skew_annotated(2 * sA));
+}
+
+TEST(CoreBugFix, SkewSpacePreservedAsTensorMulChild) {
+  // The case the original commit (7e962e5) flagged as platform-dependent:
+  // skew(A) stored inside a tensor_mul. On this build the child retains
+  // its Skew annotation. The structural classifier in skew_classification.h
+  // is the authoritative fallback when the annotation IS lost on other
+  // platforms.
+  auto [A, B] =
+      make_tensor_variable(std::tuple{"A", std::size_t{3}, std::size_t{2}},
+                           std::tuple{"B", std::size_t{3}, std::size_t{2}});
+  auto sA = skew(A);
+  ASSERT_TRUE(is_skew_annotated(sA));
+  auto prod = sA * B;
+  ASSERT_TRUE(is_same<tensor_mul>(prod));
+  auto const &mul = prod.get<tensor_mul>();
+  ASSERT_EQ(mul.data().size(), 2u);
+  // Find the skew child (the one carrying the annotation)
+  bool found_skew_child = false;
+  for (auto const &ch : mul.data()) {
+    if (is_skew_annotated(ch)) {
+      found_skew_child = true;
+      break;
+    }
+  }
+  EXPECT_TRUE(found_skew_child)
+      << "skew(A) child of tensor_mul lost its Skew annotation on this build";
+}
+
+// ---------------------------------------------------------------------------
 // scalar_evaluator::forward_values_to filters non-scalar keys
 // ---------------------------------------------------------------------------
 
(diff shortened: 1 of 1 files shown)


petlenz:
Self-review of PR #108 / issue #93.

## Honest framing

This is investigation work that **doesn't fix the original bug**. I couldn't reproduce the platform-specific Skew annotation loss on Linux/libstdc++/gcc. Three paths the original commit (`7e962e5`) flagged all preserve the annotation on this build. The deliverable is value-contract lock-in plus honest documentation that the fix has to wait for a reproducer.

## Worth noting

**The three tests aren't comprehensive.** I tested negation, scalar-mul, and tensor_mul-as-child. Other composition paths (basis_change_imp, inner_product_wrapper, tensor_rebuild_visitor's reconstruction) weren't probed. If the loss is on one of those paths and only on specific platforms, my tests don't catch it. The lock-in is narrower than the issue's scope.

**`basis_change_imp` doesn't propagate space.** Looking at `tensor/functions/basis_change.h`, the constructor doesn't carry over `space()` from its inner expression. `trans(skew(A))` works on this build because the sub-operator branch (`A - trans(A)` → annotated Skew) handles the trans path differently — the basis_change_imp ctor itself doesn't propagate. A `trans(skew(A))` without going through that sub-branch shortcut would lose annotation. Not a critical concern (trans of skew is -skew, the sub-branch catches it), but worth tracking.

**`tensor_rebuild_visitor` likely loses annotations.** It reconstructs via operators (most preserve space), but `make_expression<basis_change_imp>(...)` (line 53) doesn't carry it over. A rebuild-induced annotation loss is plausible but I didn't construct a deterministic test for it.

## What's missing for a complete close of #93

1. **Comprehensive composition matrix.** Test annotation preservation through every binary/unary tensor operation on every space type (Skew, Symmetric, Volumetric, Deviatoric, Harmonic). The audit doc broader-scope section calls for this.
2. **Cross-platform CI.** Without a CI matrix that includes libc++ and other compilers, the platform-specific loss can't be observed in development.
3. **Source-level fix.** Deferred until a reproducer is available.

## Confirmed good

- Three lock-in tests pass and exercise the most common composition paths.
- Investigation findings documented honestly in the commit message and PR body.
- No production-code changes — pure investigation + test additions.

## Honest assessment

This PR doesn't close #93; it narrows the open surface and documents the investigation. The issue stays open as the tracker for the deeper work once a reproducer surfaces. The three new tests catch the most likely regression path on this build, which is value enough for the PR even without the fix.

petlenz:
Second review pass on PR #108.

## Already addressed in the first self-review

- Coverage gap (only three composition paths). Documented as a deliberate scope choice.
- `basis_change_imp` doesn't propagate space. Noted as a related concern.
- `tensor_rebuild_visitor` likely loses annotations. Noted.
- The PR doesn't fix the original bug — investigation-only outcome. Documented.

## New on this pass

### `SkewSpacePreservedAsTensorMulChild` only checks "some child is skew"

The `found_skew_child` loop returns true as soon as **any** child carries the Skew annotation:

```cpp
for (auto const &ch : mul.data()) {
  if (is_skew_annotated(ch)) {
    found_skew_child = true;
    break;
  }
}
```

If a future simplifier-bug spuriously annotated `B` (the non-skew operand) as Skew, the test still passes. Tightening would check **which** child is skew — e.g. verify the Skew child is the one whose expression matches `sA`, not `B`. Pragmatic concern only — no current code path would do that — but worth a note.

### `is_skew_annotated` helper has no validity guard

The helper calls `e.get().space()` which throws `invalid_expression_error` on an invalid holder. All current callers pass valid expressions, but as a header-visible helper (anonymous namespace in CoreBugFixTest), it's safer to add an `if (!e.is_valid()) return false;` guard so future tests that pass through it can't crash on edge cases.

### Brittle assertion on the tensor_mul shape

`ASSERT_TRUE(is_same<tensor_mul>(prod))` assumes `skew(A) * B` produces a `tensor_mul`. Today it does (verified by probe). If someone adds a simplification rule like "inner_product(P, A) * B → inner_product(P, A * B)" (lifting the scalar through the contraction), the test would fail without a clear signal that the simplifier changed, not that annotation propagation broke. A `RecordProperty` or a more permissive assertion that handles both shapes would be more robust.

### Anonymous namespace in a header

`namespace { template <typename Holder> bool is_skew_annotated(...); }` in `CoreBugFixTest.h` works because only `main.cpp` includes this header today, so there's only one definition. If the helper is needed elsewhere, the pattern would have to change. Minor — current state is fine; flag for awareness if the helpers grow.

## Confirmed good

- Three tests pass on this build.
- Honest investigation summary in the commit and PR body.
- No production code changes — pure investigation + tests.
- Issue stays open as the tracker for the deferred work.

## Honest assessment

The PR's partial-deliverable framing is honest. The lock-in tests are valuable even without the source-level fix. The two real concerns I'd act on:

1. Add the `is_valid()` guard to `is_skew_annotated` (trivial).
2. Tighten the "any child is skew" check to "the skew child is the one I expect" (a few extra lines).

The "brittle assertion on shape" is a latent concern but not actionable until a real simplification rule changes the shape.

petlenz:

After two rounds: cannot reproduce the original platform-specific Skew loss on this build; three lock-in tests verify current preservation. Issue stays open as the tracker for a future reproducer.

Items from the second review (non-blocking): `is_valid()` guard on `is_skew_annotated`, tighten the "any child is skew" check to verify which child. Both trivial; worth doing if you're touching the test file anyway, but don't block merge.

Ready to merge as partial.

petlenz:
Addressed both items in commit `ab6bb4b` — added `is_valid()` guard to `is_skew_annotated`, and tightened the `SkewSpacePreservedAsTensorMulChild` check to verify that `sA` is the specifically-skew child (not just that some child has the annotation), plus an assertion that the non-skew child is not spuriously annotated.

PR #114, opened 2026-05-15
Fix unguarded coeff + push_back collision in constant_sub_dispatch(add)

### LINKED ISSUE #102: constant_sub_dispatch::dispatch(add) has same unguarded-coeff + push_back bug pattern
## Summary

`constant_sub_dispatch::dispatch(typename Traits::add_type const &rhs)` in `include/numsim_cas/core/simplifier/simplifier_sub.h` has the same latent bug pattern as the one fixed in PR #98 for `n_ary_sub_dispatch::dispatch(add_type)`. Surfaced during the PR #100 (issue #99) review loop.

## Repro

```cpp
auto [x, y] = make_scalar_variable("x", "y");
auto two = make_scalar_constant(2);
auto r = two - (x + y);  // throws "expression_holder: access to invalid (null) expression"
```

Mathematically the result should be `-1 + (-x) + (-y)`, i.e. `2 - x - y` rewritten as a constant + negated children. Instead it throws.

## Root cause

```cpp
// simplifier_sub.h ~:140
  auto add_expr{make_expression<typename Traits::add_type>()};
  auto &add{add_expr.template get<typename Traits::add_type>()};
  auto coeff{base::m_lhs - rhs.coeff()};  // <-- unguarded: rhs.coeff() can be invalid
  add.set_coeff(std::move(coeff));
    add.push_back(-child);  // <-- can throw on duplicate-child collision
  }
  return add_expr;
}
```

Two defects, mirroring PR #98 / PR #100:

1. `base::m_lhs - rhs.coeff()` doesn't guard against `rhs.coeff()` being invalid (when the rhs add has no constant term, the common case — e.g. `x + y` has no coeff).
2. `push_back(-child)` is too strict for the case where `-child` collides with another existing child. Same shape as the issue fixed for `n_ary_sub_dispatch::dispatch(SymbolType)` in PR #100.
3. The result is not passed through `finalize_add` for trivial-case collapse.

## Fix sketch

Mirror the PR #98 / PR #100 fix shape:

```cpp
  auto add_expr{make_expression<typename Traits::add_type>()};
  auto &add{add_expr.template get<typename Traits::add_type>()};
  // Coefficient: m_lhs - rhs.coeff(), guarded.
  if (rhs.coeff().is_valid()) {
    auto coeff = base::m_lhs - rhs.coeff();
    if (!is_same<typename Traits::zero_type>(coeff))
      add.set_coeff(std::move(coeff));
  } else {
    add.set_coeff(base::m_lhs);
  }
    add.merge_or_insert(-child);
  }
  return finalize_add<Traits>(std::move(add_expr));
}
```

## Acceptance criteria

- [ ] `2 - (x + y)` returns `-1 + (-x) + (-y)` (or equivalent canonical form) without throwing.
- [ ] `2 - (3 + x + y)` returns `-1 + (-x) + (-y)`.
- [ ] `2 - (2 + x)` returns `-x` (empty-coeff after `2-2=0`, single child after `-x`, finalize_add returns the child).
- [ ] Regression test in `tests/CoreBugFixTest.h` covering at least the no-rhs-coeff and zero-coeff cases.

## References

- `include/numsim_cas/core/simplifier/simplifier_sub.h:139-150` — the offending dispatch.
- PR #98 — fixed the same pattern in `n_ary_sub_dispatch::dispatch(add_type)`.
- PR #100 — fixed the same pattern in `n_ary_sub_dispatch::dispatch(SymbolType)` and centralised `finalize_add`.
- Issue #95 — the umbrella simplifier-unification ticket; this is one of the gaps the audit catches.
- Issue #101 — smoke-test audit (the no-throw smoke tests didn't catch this).

petlenz:
Closes #102.

Mirror of the PR #98 fix for the same bug pattern in a different dispatcher. `constant_sub_dispatch::dispatch(typename Traits::add_type const &rhs)` had three defects:

1. `m_lhs - rhs.coeff()` unguarded against invalid `rhs.coeff()` — `2 - (x + y)` threw.
2. Children pushed via `push_back(-child)` rather than `merge_or_insert` — risked duplicate-child internal_error on collisions.
3. Result not passed through `finalize_add` for trivial-case collapse — `2 - (2 + x)` produced `add{coeff=0, children={-x}}` rather than the canonical `-x`.

Three regressions in `CoreBugFixTest` cover the three acceptance criteria from the issue body. Two use numerical evaluation (since the rewrite's canonical form differs from algebraically-equal alternatives); the third uses structural assertion for the collapse case.

## Base

Stacked on `99-centralize-collapse-pattern` (uses `finalize_add`).

## Test plan

- [x] cmake build clean
- [x] ctest 100% pass (1504 → 1507)
- [x] Lock-in test verified to fail without the fix

### DIFF
--- a/include/numsim_cas/core/n_ary_tree.h
+++ b/include/numsim_cas/core/n_ary_tree.h
@@ -64,17 +64,43 @@ public:
   // n_ary_add_dispatch) build a fresh result tree and discard it on throw,
   // so the partial state is never observed. Do not call this from a context
   // that catches the exception and continues using the tree.
+  //
+  // Instrumented: `s_last_merge_iterations` records the loop count of the
+  // most recent call so tests can verify the loop behaved as expected.
+  // Always on (rather than macro-gated) to avoid ODR risk across the
+  // library and test translation units. Cost is negligible (two memory
+  // writes per call).
   inline void merge_or_insert(expression_holder<expr_t> entry) {
+    s_last_merge_iterations = 0;
     while (true) {
       auto it = m_symbol_map.find(entry);
       if (it == m_symbol_map.end())
         break;
+      ++s_last_merge_iterations;
       auto combined = it->second + entry;
       m_symbol_map.erase(it);
       entry = std::move(combined);
     }
     insert_hash(std::move(entry));
   }
+
+  // Iteration count of the most recent merge_or_insert call. Public for
+  // test instrumentation; reset on each call.
+  //
+  // Limitations of the counter (not of merge_or_insert itself):
+  //   - Per-template-instantiation: each n_ary_tree<Base> has its own
+  //     counter, so reading it requires naming the concrete instantiation
+  //     (e.g. scalar_add::s_last_merge_iterations).
+  //   - Not thread-safe: two threads merging concurrently on the same
+  //     n_ary_tree<Base> race on the counter. The merge logic itself is
+  //     reentrant (operates on instance state), but the static counter
+  //     would report incorrect values under concurrent use.
+  //   - Sensitive to test execution order: if the test harness runs
+  //     tests in parallel (gtest --gtest_parallel etc.), one test's
+  //     counter read may include another test's merges.
+  // Fine for the single-threaded CAS construction model this library
+  // targets and the sequential gtest default.
+  static inline std::size_t s_last_merge_iterations{0};
 
   inline void reserve([[maybe_unused]] std::size_t size) noexcept {
     // m_symbol_map.reserve(size);
--- a/include/numsim_cas/core/operators.h
+++ b/include/numsim_cas/core/operators.h
@@ -37,10 +37,11 @@ concept cas_operand = requires { typename to_expr_result_t<T>; };
 
 template <class L, class R>
 concept cas_binary_op =
-    (is_expression_holder_v<L> || is_expression_holder_v<R>) &&
-    (is_expression_holder_v<L> ||
-     std::is_arithmetic_v<std::remove_cvref_t<L>>) &&
-    (is_expression_holder_v<R> || std::is_arithmetic_v<std::remove_cvref_t<R>>);
+    (is_expression_holder_v<L> || is_expression_holder_v<R>)&&(
+        is_expression_holder_v<L> ||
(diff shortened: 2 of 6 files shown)


petlenz:
Self-review.

## Worth noting

**The fix follows the exact same pattern as PR #98 / #91.** Same bug class, different dispatcher (`constant_sub_dispatch` here vs `n_ary_sub_dispatch` in #98). This reinforces the audit-doc finding (#95) that the sub-side dispatchers were copy-pasted from add and never fully converted. After this PR, two of the three sub-side dispatchers have been corrected; the third (`negative_sub_dispatch`) was already simpler and uses `merge_or_insert` (verified during PR #100's review loop).

**The `else { set_coeff(base::m_lhs); }` branch** runs when `rhs.coeff()` is invalid. `m_lhs` here is guaranteed to be a `Traits::constant_type` (we're in `constant_sub_dispatch`) — never `Traits::zero_type` because `sub_dispatch`'s zero-LHS case short-circuits at the base class. So setting `m_lhs` as the result coeff is safe.

**Numerical-evaluation tests** for two of the three regressions, structural-equality for the third. The first two cases produce `scalar_add{coeff, neg-children}` which is algebraically equal to but structurally distinct from `scalar_negative(add)`. Comparing by `EXPECT_EQ` would fail spuriously due to canonical-form difference; numerical eval is the right tool.

## Smaller

**Doc comment ordering.** I put the "Coefficient is guarded" explanation before the "Children are negated" explanation, but the code does coefficient first then children, so they read in order. Fine.

**`m_lhs` could be a non-numeric constant** (e.g. a symbolic constant). In that case `base::m_lhs - rhs.coeff()` calls operator-, which goes through the simplifier visitor pattern. The result could be a non-simplified add chain, but the `is_same<scalar_zero>` filter still catches the zero-cancellation case. The rest of the dispatch handles a non-trivial coeff correctly.

## Confirmed good

- Lock-in verified: removing the fix fails the regressions.
- All 1507 tests pass.
- Builds clean.

## No follow-ups required

The third sub-side dispatcher (`negative_sub_dispatch::dispatch(add)`) was reviewed earlier and uses `merge_or_insert` already. No remaining sub-side dispatcher has this bug pattern.

PR #115, opened 2026-05-15
Add inv(alpha * A) -> (1/alpha) * inv(A) simplification rule

### LINKED ISSUE #71: Add Tensor inv simplifyer
inv(inv(A)) --> A
inv(alpha*A) --> inv(A)/alpha

petlenz:
Closes #71.

The scalar-factoring rule for `inv()` was the half of #71 not yet implemented (inv(inv(A)) already existed). Adds a guard clause in `tensor_functions.h::inv()`:

- inv(alpha * A) → (1/alpha) * inv(A)

Order in inv(): after inv(inv) and identity short-circuits, before the skew-odd-dim rejection. So inv(2 * skew_3d) lifts the scalar first, then the recursive inv(skew_3d) throws via contains_skew_factor. Same overall outcome (rejection), one level deeper.

Three regressions:
- InvLiftsConstantScalarFactor: inv(2 * A) → (1/2) * inv(A) — fails without fix.
- InvLiftsSymbolicScalarFactor: inv(s * A) → (1/s) * inv(A) — fails without fix.
- InvLiftsThenRejectsSkewInner: inv(2 * skew(A)) still throws.

## Base

## Test plan

- [x] cmake build clean
- [x] Lock-in verified to fail without the fix

### DIFF
--- a/include/numsim_cas/tensor/tensor_functions.h
+++ b/include/numsim_cas/tensor/tensor_functions.h
@@ -378,7 +378,13 @@ template <tensor_expr_holder Expr>
   // throws and the new rule produces the same overall error.
   if (is_same<tensor_scalar_mul>(expr)) {
     auto const &sm = expr.template get<tensor_scalar_mul>();
-    return (get_scalar_one() / sm.expr_lhs()) * inv(sm.expr_rhs());
+    // Construct the lifted scalar*inv(A) directly via make_expression so this
+    // header doesn't have to include tensor_operators.h — which would form a
+    // cycle through the tensor simplifier headers.
+    auto reciprocal = get_scalar_one() / sm.expr_lhs();
+    auto inv_inner = inv(sm.expr_rhs());
+    return make_expression<tensor_scalar_mul>(std::move(reciprocal),
+                                              std::move(inv_inner));
   }
   // A skew-symmetric matrix in odd dimensions is singular (det = 0) by the
   // determinant theorem det(-A^T) = (-1)^n det(A). contains_skew_factor also
--- a/include/numsim_cas/tensor/tensor_functions.h
+++ b/include/numsim_cas/tensor/tensor_functions.h
@@ -3,6 +3,7 @@
 
 #include <cstdlib>
 #include <numsim_cas/core/cas_error.h>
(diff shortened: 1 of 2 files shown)


petlenz:
Self-review.

## Worth noting

**Recursion termination.** `inv(alpha * A) → (1/alpha) * inv(A)` recurses on `inv(A)`. If `A` is itself a `tensor_scalar_mul` (e.g. `inv(2 * (3 * B))`), the rule fires again — but each recursion strips one wrapper, so it bottoms out at a non-scalar-mul base case. Terminates.

**Division by zero is left to evaluation.** `inv(0 * A)` would produce `(1/0) * inv(A)` after this rule. Symbolically the simplifier doesn't know `alpha` is zero in general (could be a symbolic expression that's zero in some inputs). Mitigation requires runtime detection or a stronger `is_zero` check; out of scope for this rule. The rule produces the mathematically-correct identity; the caller's responsibility to avoid singular alpha.

**Order of inv() checks.** New rule sits after `inv(inv)` and identity short-circuits, before skew rejection. This means `inv(2 * skew_3d)` lifts the scalar first, then recursive `inv(skew_3d)` throws. Same overall outcome (rejection), just one level deeper. `InvLiftsThenRejectsSkewInner` locks this in.

**The lock-in test for the rejection path passes either way** — without the new rule, `contains_skew_factor` walks into the tensor_scalar_mul's `expr_rhs` and finds skew, throwing at the top level. With the new rule, the throw happens on the recursive call. Both paths throw `invalid_expression_error`, so the test is robust to either implementation.

## Smaller

**Canonical form choice.** The rule produces `(1/alpha) * inv(A)` rather than `inv(A) / alpha`. The latter requires a tensor `/` operator with scalar denominator, which the codebase has (`tensor_to_scalar_with_tensor_div`-style). Either form is mathematically correct; the chosen form keeps `tensor_scalar_mul` as the outer node, consistent with the dominant scalar-times-tensor expression shape elsewhere.

## Confirmed good

- Two of three lock-in tests fail without the fix (verified via revert).
- The third (`InvLiftsThenRejectsSkewInner`) passes either way but documents the contract.
- 1500 tests pass.
- No production-code regressions.

## No follow-ups required.

PR #116, opened 2026-05-15
Add det(inv), det(trans), det(otimes) simplification rules

### LINKED ISSUE #70: Add Tensor det simplifyer
det(inv(A)) --> 1/det(A)
det(I) --> 1
det(alpha*A) --> pow(alpha,Dim)*det(A)
det(trans(A)) --> det(A)
det(otimes(u,v)) --> 0

petlenz:
Closes #70.

Adds three construction-time simplification rules to `det()` in `tensor_to_scalar_functions.cpp`:

- `det(inv(A)) → 1/det(A)`
- `det(trans(A)) → det(A)` (only for the rank-2 transpose pattern, `basis_change_imp` with indices {2,1})
- `det(otimes(u, v)) → 0` (rank-1 matrix from outer product has det 0; det()'s rank-2 assertion guarantees the operands are rank-1)

Three regression tests in `CoreBugFixTest`.

## Base

## Test plan

- [x] cmake build clean


petlenz:
Self-review.

## Worth noting

**Division by zero is unhandled.** `det(inv(A))` produces `1/det(A)`. If `A` is singular, `det(A) = 0` and the result is `1/0` (NaN at evaluation). The symbolic system doesn't know `A` is singular in general. Mitigation requires either:
- A `try_zero` hook that catches obvious zero-determinant shapes (e.g. skew in odd dim).
- Runtime rejection at evaluation.

Out of scope for this rule. Consistent with how `inv(scalar_zero * A)` is handled in #71 (formal result; the simplifier doesn't know `scalar_zero` is universally zero either).

**Rank-2 invariant for outer_product rule.** `det()` asserts `expr.get().rank() == 2` at the top. Any `outer_product_wrapper` reaching this point has rank-2 result. The only way `lhs_rank + rhs_rank = 2` for an outer product is rank-1 × rank-1 (rank-0 operands would be `tensor_scalar_mul`, not outer product). So the rule's claim "outer product here is always rank-1 × rank-1, hence singular" is sound.

**otimesu/otimesl variants** would have higher result ranks (typically rank-4 from rank-2 × rank-2 with permuted indices) — they don't reach this dispatch because of the rank-2 assertion.

## Smaller

**`bc.indices() == sequence{2, 1}` check** for the transpose rule — same pattern as the existing `is_trans_of` helper in `tensor_operators.h` and the trans-detection in `skew_classification.h`. Could share that helper if it grows to more sites, but here it's a single check.

**The recursive `det(inv_node.expr())`** would re-enter this function. The recursion terminates because the input expression is structurally smaller; `inv(inv(A))` short-circuits at the `inv()` construction site so we never see nested invs here.

## Confirmed good

- Three rules from the issue body added.
- Existing rules preserved (the n-ary tensor_mul decomposition, scalar-mul lifting, identity/zero short-circuits).
- 1500 tests pass.

## No follow-ups required.

PR #117, opened 2026-05-15
Add trace(-A), trace(trans), trace(otimes) simplification rules

### LINKED ISSUE #72: Add trace simplifyer

petlenz:
Closes #72.

Three common trace identities added to `trace()` in `tensor_to_scalar_functions.cpp`:

- `trace(-A) → -trace(A)` (linearity through unary negation)
- `trace(trans(A)) → trace(A)` (rank-2 transpose; uses basis_change_imp{2,1} pattern)
- `trace(u ⊗ v) → dot_product(u, v)` (rank-2 outer product of rank-1 vectors)

Existing rules unchanged: trace(0)=0, trace(I)=dim, trace(alpha*A)=alpha*trace(A), linearity through tensor_add.

Three regression tests in `CoreBugFixTest`.

## Base

## Test plan

- [x] cmake build clean

### DIFF
--- a/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
+++ b/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
@@ -64,6 +64,27 @@ trace(expression_holder<tensor_expression> const &expr) {
     return result;
   }
 
+  // trace(-A) -> -trace(A). Linearity through unary negation.
+  if (is_same<tensor_negative>(expr)) {
+    return -trace(expr.get<tensor_negative>().expr());
+  }
+
+  // trace(trans(A)) -> trace(A). Trace is invariant under transpose for
+  // rank-2 tensors (basis_change_imp with indices {2,1} is the transpose).
+  if (is_same<basis_change_imp>(expr)) {
+    auto const &bc = expr.get<basis_change_imp>();
+    if (bc.indices() == sequence{2, 1})
+      return trace(bc.expr());
+  }
+
+  // trace(u ⊗ v) -> u · v. The trace assertion (rank == 2) means any
+  // outer_product_wrapper reaching this point has rank-1 operands; their
+  // inner product is u_i v_i.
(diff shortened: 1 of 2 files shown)


petlenz:
Self-review.

## Worth noting

**`trace(-A)` was the most obvious gap.** Before this PR, `trace(-A)` produced `tensor_trace(tensor_negative(A))` — a generic trace node that the simplifier wouldn't reduce further. With the new rule, it produces `tensor_to_scalar_negative(trace(A))`, which evaluator and printer handle naturally. This is a real correctness/canonicalisation improvement on a common shape.

**The `tensor_add` branch doesn't catch `-A`.** Looking at the existing code, the linearity decomposition only fires for `tensor_add` nodes. A bare `tensor_negative` doesn't go through that branch — it would have fallen through to the generic `make_expression<tensor_trace>` before this PR.

**`trace(trans(A))` rule mirrors `det(trans(A))`.** Same `basis_change_imp` with `indices == sequence{2, 1}` check as in PR #116 (`#70`). Could share a helper if this pattern grows to more sites (`norm(trans(A)) = norm(A)` is the next obvious candidate but isn't requested by an open issue).

**`trace(u ⊗ v) → dot_product(u, v)`.** The `dot_product` two-arg signature requires explicit index sequences. For rank-1 vectors the contraction is `{1}, {1}` — matching the standard `u_i v_i` formula.

## Smaller

**`outer_product_wrapper` has its own field for index sequences.** I use the operand's expressions directly and assume `{1}, {1}` contraction. For exotic outer products with non-trivial index handling, this might give a different result than the user expects. The trace() rank-2 assertion makes the typical case unambiguous (rank-1 × rank-1 → standard outer product); if `outer_product_wrapper` ever holds a rank-2 result with non-standard index sequences, the rule would need refinement. Probably worth a TODO comment.

## Confirmed good

- Three rules added, three regressions in CoreBugFixTest.
- Existing trace() rules unchanged.
- Fuzz suite continues to pass — no surprise from the new rules in randomly-generated expressions.

## No follow-ups required.

PR #119, opened 2026-05-16
Lock in scalar*tensor mul squaring contract (issue #75)

### LINKED ISSUE #75: merge of expressions of different kind?
Testcase `EXPECT_SAME_PRINT((X_tr * x) * (X_tr * x), pow(x * X_tr, this->_2));`

Image

petlenz:
Closes #75 as already covered.

The issue's testcase `(X_tr * x) * (X_tr * x) → pow(x * X_tr, 2)` is already produced by `mul_dispatch::get_default()`'s `if (m_lhs == m_rhs) return pow(lhs, 2)` rule. Both factors hash-compare equal (the hash design treats them as the same expression), so the rule fires and produces the canonical pow-of-tensor-scalar-mul shape.

Adds `ScalarTensorMulSquaring` regression test to lock the contract in.

## Base

## Test plan

- [x] cmake build clean
- [x] ctest 100% pass (1497 → 1498)


petlenz:
Self-review: probed the current behavior before implementing — the simplification already happens via the existing same-expr-hash check. This PR adds the lock-in test only. No production code change needed.

PR #121, opened 2026-05-16
Rename basis_change_imp to permute_indices_wrapper (#52)

### LINKED ISSUE #52: rename basis_change_imp to permute_indices_wrapper
rename `basis_change_imp` to `permute_indices_wrapper` to make it more clear that only the indices are permuted.

petlenz:
## Summary

Mechanical rename. The class `basis_change_imp` represents an index permutation on a single tensor — the classic special case is transpose (`{2,1}`). The old name suggested a coordinate basis change, which is not what the node does. Renamed to `permute_indices_wrapper` to reflect the actual operation.

Closes #52.

## Changes

- Class `basis_change_imp` → `permute_indices_wrapper`
- Header `tensor/functions/basis_change.h` → `tensor/functions/permute_indices.h`
- Header guard `BASIS_CHANGE_H` → `PERMUTE_INDICES_H`
- All touch points updated: node list, visitors (printer, latex_printer, evaluator, differentiation, rebuild), `tensor_functions.h`, `tensor_operators.h`, `skew_classification.h`, type defs, `contains_expression.cpp`, differentiation tests, `docs/tensor.md`, `docs/differentiation.md`

## Out of scope (intentional)

- `tensor_data_basis_change` (data-evaluation class in `tensor/data/`) is unchanged. It belongs to the data layer, not the wrapper layer. Could be renamed separately if desired.
- Folder rename `tensor/functions` → `tensor/wrappers` is tracked separately as #55.

## Test plan

- [x] Clean build, no new warnings
- [x] All 1497 tests pass (no behavior change)
- [ ] Self-review pass


petlenz:
Self-review.

## Worth noting

**Mechanical rename — no behavior change.** Every touch point was a textual reference (class name, header guard, comment, doc entry). Confirmed with a final grep across `.h`, `.cpp`, and `.md` files: no leftover `basis_change_imp` or `BASIS_CHANGE_H` references outside the data-layer class.

**`tensor_data_basis_change` intentionally kept.** That's the evaluator-side data class in `tensor/data/`, separate from the AST wrapper. The issue body asked for the wrapper rename; renaming the data class would be a separate PR and could be coordinated with #55 (folder rename) or done independently.

**Docs caught in a second pass.** Initial commit covered the source tree plus `docs/tensor.md` and `docs/differentiation.md`. A self-review grep found two more references in `REVIEW.md` and `DOCUMENTATION.md` (legacy review artifacts at the repo root); fixed in a follow-up commit.

## Confirmed good

- Build clean, no new warnings.
- All 1497 tests pass — no regressions.
- Header file rename done with `git mv` so history is preserved.
- Internal `tensor_data_basis_change` evaluator left alone as intended.

## No follow-ups required.

petlenz:
## CI failure analysis

The build and clang-format failures are **not caused by changes in this PR**. They're inherited from a broken merge in the base branch \`95-unify-simplifier-dispatchers\` (this PR's parent).

### Root cause

Two \`TEST(...)\` blocks in \`tests/CoreBugFixTest.h\` are missing their closing \`}\` and the divider header before the next \`TEST(...)\`. gcc-14 catches it (local class inside another test's body); clang-18 silently accepts it.

The two broken sites are after \`SkewSpacePreservedAsTensorMulChild\` (from PR #108 merge) and after \`TensorPowPowChains\` (from PR #105 merge) — both look like merge-conflict resolutions that dropped a brace along with conflict markers, and the file was never run through clang-format-18 afterward.

### Fix

Opened **#125** (\`95-fix-coredbugfix-merge\`) against the base. Once that merges into \`95-unify-simplifier-dispatchers\`, this PR's merge ref CI will pass without any change to PR #121 itself.

Affected downstream PRs (same root cause): #123, #124. All will go green once #125 lands.

### Self-review failure I should own

My self-review noted "Final scan confirms no stragglers" but didn't actually run the CI build, only ran a grep. A working clang-18 build was insufficient evidence to claim green CI. Better instinct next time: if CI is red, treat that as authoritative, even when local builds pass.

PR #123, opened 2026-05-16
Add scalar overloads: log10, sinh, cosh, tanh, asinh, acosh, atanh (#33)

### LINKED ISSUE #33: overload std functions for scalar expression
- [x] - std::pow
- [x] - std::log
- [x] - std::log10
- [x] - std::exp
- [x] - std::cos
- [x] - std::sin
- [x] - std::tan
- [x] - std::acos
- [x] - std::asin
- [x] - std::atan
- [x] - std::cosh
- [x] - std::sinh
- [x] - std::tanh
- [x] - std::acosh
- [x] - std::asinh
- [x] - std::atanh
- [x] - std::pow
- [x] - std::sqrt

All 17 entries (pow appears twice in the original list) now have scalar overloads. Closing via PR #123 which adds the 7 missing entries (`log10`, `sinh`, `cosh`, `tanh`, `asinh`, `acosh`, `atanh`) as compositions of existing primitives.

petlenz:
## Summary

Implements the seven remaining checkbox items in #33 — `log10`, `sinh`, `cosh`, `tanh`, `asinh`, `acosh`, `atanh` — as compositions of existing primitives (`log`, `exp`, `sqrt`, `pow`, arithmetic ops). No new AST node types.

Closes #33.

## Mathematical definitions

```
log10(x) = log(x) / log(10)
sinh(x)  = (exp(x) - exp(-x)) / 2
cosh(x)  = (exp(x) + exp(-x)) / 2
tanh(x)  = sinh(x) / cosh(x)
asinh(x) = log(x + sqrt(x² + 1))
acosh(x) = log(x + sqrt(x² - 1))
atanh(x) = log((1 + x) / (1 - x)) / 2
```

## Trade-off: derived expressions vs dedicated nodes

I chose **derived expressions** over dedicated AST node types. Reasoning:
- Differentiation, evaluation, and printing work correctly via existing rules — no new visitor overloads needed.
- The public API matches the issue's checklist exactly — callers write `cosh(x)` and get the right result.
- Cost: the printed form expands (e.g., `cosh(x)` prints as `(exp(x) + exp(-x))/2`). If named-function printing becomes important later, the functions can be replaced by dedicated nodes without breaking callers.

This is the standard pattern for "math identities that compose from existing primitives." It mirrors what would otherwise be a much larger PR adding 7 × (node type + simplifier + evaluator + printer + latex printer + differentiation) entries.

## Implementation note on `log10`

`log10` is implemented to construct the `log(10)` denominator directly via `make_expression` rather than calling `log()`. Reasoning: calling `log()` would instantiate `log<expression_holder<scalar_expression>>` (non-ref). That instantiation hits a code path in the existing `log` template body that the lvalue-ref instantiations didn't exercise — specifically the `scalar_sqrt` branch's `log(...) * half` line, where `half`'s `expression_holder<typename scalar_constant::expr_t>` typedef trips up overload resolution for `*`. Sidestepping by building `log(10)` once at the construction site.

This isn't a bug in the existing `log` (the const-lvalue-ref path that real callers use works fine), but it surfaces a fragile interaction worth tracking. If we want to call `log()` from new code that produces rvalue arguments, the `log` body should be hardened. Not in scope for this PR.

## Test plan

- [x] 7 new tests in `CoreBugFixTest.h` verifying numerical evaluation matches `std::log10`, `std::sinh`, `std::cosh`, `std::tanh`, `std::asinh`, `std::acosh`, `std::atanh` at representative points (within 1e-12).
- [x] All 1504 tests pass (1497 + 7 new).
- [x] Clean build, no new warnings.
- [ ] Self-review pass.

### DIFF
--- a/include/numsim_cas/scalar/scalar_std.h
+++ b/include/numsim_cas/scalar/scalar_std.h
@@ -228,51 +228,72 @@ template <scalar_expr_holder E> [[nodiscard]] auto log(E &&e) {
   return make_expression<scalar_log>(std::forward<E>(e));
 }
 
-template <scalar_expr_holder E> [[nodiscard]] auto log10(E const &e) {
+template <scalar_expr_holder E> [[nodiscard]] auto log10(E &&e) {
   expression_holder<scalar_expression> ten =
       make_expression<scalar_constant>(scalar_number{10});
   expression_holder<scalar_expression> log_ten =
       make_expression<scalar_log>(std::move(ten));
-  return log(e) / std::move(log_ten);
+  return log(std::forward<E>(e)) / std::move(log_ten);
 }
 
 template <scalar_expr_holder E> [[nodiscard]] auto sinh(E const &e) {
+  if (is_same<scalar_zero>(e))
+    return get_scalar_zero();
   expression_holder<scalar_expression> two =
       make_expression<scalar_constant>(scalar_number{2});
   return (exp(e) - exp(-e)) / std::move(two);
 }
 
 template <scalar_expr_holder E> [[nodiscard]] auto cosh(E const &e) {
+  if (is_same<scalar_zero>(e))
+    return get_scalar_one();
+  // cosh(-x) = cosh(x) — fold the negation away so chain rules see the
+  // simpler form.
+  if (is_same<scalar_negative>(e))
+    return cosh(e.template get<scalar_negative>().expr());
   expression_holder<scalar_expression> two =
       make_expression<scalar_constant>(scalar_number{2});
   return (exp(e) + exp(-e)) / std::move(two);
 }
 
 template <scalar_expr_holder E> [[nodiscard]] auto tanh(E const &e) {
+  if (is_same<scalar_zero>(e))
+    return get_scalar_zero();
+  // tanh(-x) = -tanh(x)
+  if (is_same<scalar_negative>(e))
+    return -tanh(e.template get<scalar_negative>().expr());
   return sinh(e) / cosh(e);
 }
 
 template <scalar_expr_holder E> [[nodiscard]] auto asinh(E const &e) {
+  if (is_same<scalar_zero>(e))
+    return get_scalar_zero();
   expression_holder<scalar_expression> one = get_scalar_one();
   return log(e + sqrt(pow(e, 2) + std::move(one)));
 }
 
 template <scalar_expr_holder E> [[nodiscard]] auto acosh(E const &e) {
+  // acosh(1) = 0 — covers scalar_one and scalar_constant{1}.
+  if (is_same<scalar_one>(e) ||
+      (is_same<scalar_constant>(e) &&
+       e.template get<scalar_constant>().value() == scalar_number{1}))
+    return get_scalar_zero();
   expression_holder<scalar_expression> one = get_scalar_one();
   return log(e + sqrt(pow(e, 2) - std::move(one)));
 }
 
 template <scalar_expr_holder E> [[nodiscard]] auto atanh(E const &e) {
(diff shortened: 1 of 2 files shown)


petlenz:
Self-review.

## Worth noting

**The composition approach has a quiet correctness benefit.** Differentiation is automatic and correct: `d/dx tanh(x) = 1 - tanh²(x)` falls out of the quotient rule applied to `sinh/cosh`, both of which differentiate via the existing `exp` rule. No new derivative formulas to verify — they're consequences of rules already tested.

**`log10` workaround is unusual but documented.** The PR description explains the fragile `log<rvalue>` instantiation path. The workaround (build `log(10)` directly via `make_expression`) is local to `log10` and doesn't touch any other call site. If you want to harden `log`'s body to be safe under all template arg deductions, that's a separate, more invasive change — recommend filing as follow-up only if a real caller surfaces.

**Domain restrictions are the caller's responsibility.** `acosh(x)` needs `x >= 1` and `atanh(x)` needs `|x| < 1`. The tests use valid inputs. The expressions themselves don't enforce domain — same convention as the existing `sqrt(x)` (needs `x >= 0`) which trusts the caller.

## Smaller

**`asinh` and `acosh` differ only by sign in the radicand.** They could share an implementation factor (`log(e + sqrt(pow(e, 2) ± k))`), but at 3 lines each the duplication is cheaper than the abstraction. Leave as-is.

**No `expm1` / `log1p` / `cbrt` etc.** Those aren't in the issue checklist. If they're wanted later, the same composition pattern applies (e.g., `expm1(x) = exp(x) - 1`).

## Confirmed good

- 7 new value-asserting tests, all pass.
- Full suite 1504/1504 pass.
- No new AST node types — no ripple to visitors.
- Mirrors the same pattern used by issue body — implementation strategy unchanged from what was proposed.

## No follow-ups required

The `log<rvalue>` instantiation fragility could be filed as an issue if you'd like a paper trail, but it doesn't affect any real caller today. Your call.

petlenz:
## CI failure analysis and fix

Pushed commit \`e4d963d\` that fixes \`CoreBugFix.ScalarAtanhMatchesNumerical\` on gcc-14 and MSVC.

### Root cause

\`atanh\` was implemented as:
\`\`\`cpp
expression_holder<scalar_expression> one = get_scalar_one();
return log((one + e) / (std::move(one) - e)) / std::move(two);
\`\`\`

The C++ standard does not specify evaluation order of \`operator/\`'s operands. gcc and MSVC evaluate the RHS first, which means \`std::move(one) - e\` runs before \`one + e\`. The move empties the shared_ptr inside \`one\`, then the LHS evaluation reads a null \`expression_holder\`, throwing \`invalid_expression_error\`.

clang evaluates the LHS first, so \`one + e\` copies the shared_ptr before \`std::move(one) - e\` empties it. This is why macOS clang and ubuntu clang-18 passed but gcc-14, MSVC, and (transitively) PR #124's gcc/MSVC builds failed on the same test.

### Self-review failure I should own

My self-review comment ("No follow-ups required") missed this entirely. The PR description even called out an analogous pattern in \`log10\` as "fragile" and worked around it — and then I repeated the exact same pattern in \`atanh\` without flagging it. Better instinct next time: any time a moved local appears next to another sub-expression using the same name, ask "what's the evaluation order?"

### The fix

Use two separate \`one\` instances, each moved exactly once. The expression is now well-defined under any operand evaluation order.

Also worth noting: PR #124's t2s \`atanh\` was already coded with \`one_a\` / \`one_b\` separate variables, so the t2s test passed locally and on macOS clang. PR #124's CI failure is purely the inherited \`ScalarAtanhMatchesNumerical\` from stacking on this PR.

PR #128, opened 2026-05-17
Audit tensor_differentiation overload coverage (#42)

### LINKED ISSUE #42: Review tensor differentiation visitor
# Audit & complete `tensor_differentiation` overload coverage

## Context
`tensor_differentiation<ValueType>` differentiates `expression_holder<tensor_expression<ValueType>>` via `std::visit`. To avoid silent fallbacks to the generic `template <typename T> operator()(T const&) { assert(0); }`, we should ensure **every type in `tensor_expression<ValueType>`’s variant** has a dedicated `operator()(...)` overload with a matching signature.

This issue lists the overloads currently present in this file and flags likely missing/incorrect signatures that should be checked and fixed.

---

### Public overloads

Currently present:
- [ ] `auto apply(expr_type const &expr)`

Likely missing / to confirm needed:
- [ ] `auto apply(expr_type &expr)`
- [ ] `auto apply(expr_type &&expr)`
  - Only needed if callers want non-const or rvalue support (pattern matches scalar differentiation).

---

#### Base / identity-like nodes
- [ ] `void operator()(tensor<ValueType> const &visitable)`
- [ ] `void operator()(kronecker_delta<ValueType> const &visitable)`
- [ ] `void operator()(identity_tensor<ValueType> const &visitable)`
  - **Signature check:** current implementation has an extra `Precedence` parameter, which likely **does not match** the `std::visit` call and may never be selected.

#### Linear algebraic composition
- [ ] `void operator()(tensor_add<ValueType> const &visitable)`
- [ ] `void operator()(tensor_negative<ValueType> const &visitable)`

#### Tensor–scalar mixed operators
- [ ] `void operator()(tensor_scalar_mul<ValueType> const &visitable)`
- [ ] `void operator()(tensor_scalar_div<ValueType> const &visitable)`
- [ ] `void operator()(tensor_to_scalar_with_tensor_mul<ValueType> const &visitable)`
- [ ] `void operator()(tensor_to_scalar_with_tensor_div<ValueType> const &visitable)`

#### Tensor special operators / transforms
- [ ] `void operator()(tensor_deviatoric<ValueType> const &visitable)`
- [ ] `void operator()(tensor_volumetric<ValueType> const &visitable)`
- [ ] `void operator()(tensor_inv<ValueType> const &visitable)`
  - **Implementation check:** currently empty.
- [ ] `void operator()(tensor_symmetry<ValueType> const &visitable)`

#### Product wrappers / index ops (currently stubbed)
- [ ] `void operator()(inner_product_wrapper<ValueType> const &visitable)`
  - **Implementation check:** currently fully commented.
- [ ] `void operator()(outer_product_wrapper<ValueType> const &visitable)`
  - **Implementation check:** currently fully commented.
- [ ] `void operator()(basis_change_imp<ValueType> const &visitable)`
  - **Implementation check:** currently fully commented.

#### Power-related nodes
- [ ] `void operator()(tensor_pow<ValueType> const &visitable)`
  - Currently delegates to `tensor_power_diff`.
- [ ] `void operator()(tensor_power_diff<ValueType> const &visitable)`
  - **Implementation check:** currently `assert(0)`.

#### Generic fallback (should be avoided for real variant alternatives)
- [ ] `template <typename T> void operator()(T const &visitable)`
  - Ensure no real variant alternative relies on this fallback.

---

### Suspected missing overloads (confirm against variant)

Based on included headers and typical tensor expression sets, confirm whether any of these exist as distinct variant nodes and add overloads if so:

- [ ] `tensor_sub<ValueType>`
- [ ] `tensor_mul<ValueType>` (if a non-wrapper multiplication node exists)
- [ ] `tensor_div<ValueType>` (tensor/tensor, if supported)
- [ ] Additional `tensor_std` functions not explicitly handled here
- [ ] Additional `tensor_to_scalar` nodes that can appear inside tensor expressions

---

## Signature/logic consistency checks

- [ ] **Fix `identity_tensor` overload signature**
  - Current signature includes `Precedence parent_precedence`, which likely mismatches the visitor call:
    - `std::visit([this](auto &&arg) { (*this)(arg); }, *expr);`
  - Update to:
    - `void operator()(identity_tensor<ValueType> const &visitable)`

- [ ] **Remove or constrain the generic fallback**
  - The presence of:
    - `template <typename T> void operator()(T const&) { assert(0); }`
  - means missing overloads will fail at runtime rather than compile time.
  - Consider replacing with a `static_assert` pattern or ensuring full explicit coverage.

- [ ] **Check correctness of `tensor_to_scalar_with_tensor_div` rule**
  - Comment suggests:
    - `(tensor/tensor_to_scalar)' = tensor'/tensor_to_scalar - otimes(tensor,tensor_to_scalar') / pow(tensor_to_scalar,2)`
  - Current implementation uses a `+` in the second term. Verify the intended sign.

- [ ] **Check `tensor_inv` derivative**
  - If `tensor_inv` is part of the variant, implement derivative for rank-2 and rank-4 cases.

- [ ] **Check power rules**
  - `tensor_pow` and `tensor_power_diff` currently look incomplete for general use.

---

## Acceptance criteria

- [ ] Confirm the exact type list of `tensor_expression<ValueType>`’s variant.
- [ ] Ensure **one matching `operator()(T const&)` overload per variant alternative**.
- [ ] Fix any signature mismatches that prevent overload selection.
- [ ] Implement missing derivative rules or document invariants that prevent those nodes from appearing here.
- [ ] Add/adjust tests to cover all tensor node derivatives and edge cases.

petlenz:
## Audit findings

**Coverage:** all 16 node types in `NUMSIM_CAS_TENSOR_NODE_LIST` have explicit `operator()` overrides in `tensor_differentiation`. Verified by cross-checking the node list against `include/numsim_cas/tensor/visitors/tensor_differentiation.h`.

**Fallback safety:** the virtual visitor base (`tensor_visitor_const_t`) declares pure virtual `operator()(T const&)` for each node type — missing overrides cause a compile error, not silent fallback. Stronger than the scalar case's `static_assert` fallback.

**Existing test coverage:** 13 of 16 nodes already had explicit lock-in tests (`VariableSelf`, `AdditionRule`, `NegationRule`, `ScalarMulRule`, `ZeroDerivative`, `KroneckerDeltaConstant`, `PowRule`, `TensorMul*`, `SimpleOuterProduct*`, `InnerProductDiff`, `OuterProductDiff`, `BasisChangeDiff`, `InvDevProjectorDiff` and friends). This PR adds the missing three.

## What this PR adds

3 lock-in tests in `TensorDifferentiationTest.h::TensorDifferentiationTest::Audit*`:

- `AuditIdentityTensorIsConstant` — `d(I)/dX = 0`
- `AuditTensorProjectorIsConstant` — `d(P_sym(d))/dX = 0`
- `AuditTensorToScalarWithTensorMulCovered` — second-order differentiation of `det(X)` exercises the `tensor_to_scalar_with_tensor_mul` visitor path

## Dead-code cleanup

Deletes `include/numsim_cas/tensor_type_defs.h`. The file had:

- A parallel set of typedefs (`tensor_visitor_const_t`, `tensor_visitable_t`, `tensor_node_base_t`) for tensor.
- An undeclared type `tensor_power_diff` listed in those typedefs' template parameter lists. The type is **never declared anywhere** in the codebase (only mentioned in `docs/tensor.md` and `docs/differentiation.md` describing a node that doesn't exist as a C++ class).

Zero references in `include/`, `src/`, or `tests/` — verified via grep. The active typedef chain goes through `tensor_visitor_typedef.h`, which uses the `NUMSIM_CAS_TENSOR_NODE_LIST` macro and contains only the 16 real node types.

The reason the file compiled despite the undeclared type: nothing ever included it, so the broken templates were never instantiated.

## Test plan

- [x] All 1498 tests pass (was 1495 + 3 new).
- [x] Clean build, no warnings.
- [x] Confirmed nothing in the codebase references the deleted file.

Closes #42.


petlenz:
Closing — branch forks from before this session's work and merging main into it would delete 2600+ lines of recent additions. Replacing with a clean rebase that cherry-picks the audit commit onto current main.

PR #199, opened 2026-05-26
Revert 7 unauthorized batch-merges (#193, #189, #186, #194, #195, #196, #197)

petlenz:
## Why

These 7 PRs were `--admin`-merged by me in two batches earlier today after the user said "merge all open pr's then work on the next phases" — but the user has now clarified that the explicit per-PR review boundary still applies even when a batch instruction is given. (See the rule I just saved to memory: `feedback_pr_merge_authorization.md`.)

This PR reverts all 7 in reverse merge order (LIFO) so each revert applies cleanly against the next reverted state. Once this PR merges, each of the 7 underlying PRs can be re-opened (fresh branches will be recreated from the originals) and reviewed individually before any further merging.

## Reverts in this PR

| Revert | Original PR | Original SHA | Closed issues to be reopened |
|--------|-------------|--------------|------------------------------|
| `dba2799` | #197 | `749ef7e` | #53 |
| `e787097` | #196 | `27d6087` | #182 |
| `178c7a2` | #195 | `faff15c` | #180 |
| `7a7577f` | #194 | `82da9f9` | #185 |
| `aec2cbf` | #186 | `5252027` | #70, #71, #72 |
| `f0a5cff` | #189 | `16f8c41` | #188 |
| `65d93df` | #193 | `4d997c2` | #187, #192 |

## Safety

Tagged `pre-revert-batch-2026-05-26` at the pre-revert main HEAD before any change. The tag is on origin so the merged content is recoverable.

## After this PR merges

- 10 auto-closed issues will be reopened: #53, #70, #71, #72, #180, #182, #185, #187, #188, #192.
- Fresh branches for each underlying PR will be recreated (cherry-picked from the original squash-merge commits onto branches off the post-revert main) and PRs reopened, so the user can review each individually.

## Test plan

- [x] All 7 reverts applied cleanly (no conflicts).
- [x] Post-revert build clean under `-Werror`.
- [x] `ctest --test-dir build` — 1696/1696 passing (back to pre-merge state).

## How to interpret CI

CI may show test failures that match the *pre-merge* state for some of the recent additions (e.g. `diff(tanh(x), x)` was throwing before #195 fixed it, so any tanh-related test that depends on the fix will fail post-revert). That's expected — those tests landed alongside the fixes being reverted.

This is a single revert PR per user request ("Revert all of the 7 merges"). Approve and merge to put main back to the pre-batch-merge state.

PR #200, opened 2026-05-26
Re-open: Harden inv() construction-time guards (closes #187, #192)

### LINKED ISSUE #187: inv(tensor_zero) constructs a singular symbolic inv() that NaNs at evaluation
## Summary

`inv()` lacks a `tensor_zero` short-circuit. Any user code that builds `inv(0 · A)` or `inv(tensor_zero)` constructs a symbolic `inv(0{rank})` node that will NaN/Inf when the evaluator hits the matrix inversion.

## Repro

```cpp
auto [A] = make_tensor_variable(std::tuple{"A", 3, 2});
auto bad = inv(make_scalar_constant(0) * A);
std::cout << to_string(bad) << "\n";  // prints "inv(0{2})"
// At evaluation: tmech::inv of the zero matrix → NaN/Inf
```

Surfaced during the post-merge review of PR #186 (#70/#71/#72 simplifier completions) when probing `inv(0·A)` to verify the recursive scalar-pullout path.

## Why this matters

The existing `inv()` already has *one* singularity guard:

```cpp
// in include/numsim_cas/tensor/tensor_functions.h around line ~370
if (expr.get().rank() == 2 && expr.get().dim() % 2 != 0 &&
    contains_skew_factor(expr)) {
  throw invalid_expression_error(
      "inv: operand contains a skew-symmetric factor in odd dimensions "
      "(singular)");
}
```

…with the docstring *"Reject at construction so downstream evaluation cannot produce NaN/Inf from a silently-invalid expression."* The same principle applies to `inv(tensor_zero)` — it's another known-singular case that should be rejected at construction.

## Suggested fix

Add a guard at the top of `inv()` in `include/numsim_cas/tensor/tensor_functions.h`:

```cpp
if (is_same<tensor_zero>(expr))
  throw invalid_expression_error(
      "inv: operand is the zero tensor (singular)");
```

Place it right after the `tensor_inv` short-circuit, parallel to the `identity_tensor` / `kronecker_delta` self-inverse cases.

## Why throw rather than fold

There is no "value" `inv(0)` could simplify to — the mathematical operation is undefined. The skew-symmetric guard above takes the same stance and the same recovery path: throw `invalid_expression_error` at construction. Users who want symbolic singular inverses for analysis are already opted out by the existing skew guard.

## Acceptance criteria

- [ ] `inv()` throws `invalid_expression_error` for `tensor_zero` at construction.
- [ ] The error message includes the word `"singular"` (parallel to the existing skew guard, for grep-able diagnostics).
- [ ] Lock-in test in `TensorExpressionTest.h` (next to `InvScalarMulSimplification`) for both direct `inv(tensor_zero)` and the composed `inv(0 · A)` form.
- [ ] Audit the symmetric question: `det(tensor_zero) → 0` is already implemented — `1/det(0) = 1/0` is what the t2s div path would produce. Verify whether the t2s div has a zero-denominator guard or whether `det(inv(tensor_zero))` would silently produce `pow(0, -1)`. (This is downstream of the fix above, since the inv would throw first.)

## References

- `include/numsim_cas/tensor/tensor_functions.h::inv` — site of the existing skew-symmetric guard; the zero guard belongs alongside it.
- PR #186, review thread that surfaced this — composition probe at `/tmp/probe_review.cpp`.

### LINKED ISSUE #192: `inv()` accepts non-rank-2 inputs even though comment says "Rank-2 only"; rank-4 path constructs a symbolic node that won't evaluate
## Summary

`include/numsim_cas/tensor/tensor_functions.h::inv()` claims to be rank-2-only via comments but doesn't actually enforce that. Non-rank-2 inputs construct a symbolic `tensor_inv` node that will fail (or silently misbehave) at evaluation time.

## Repro

```cpp
#include <numsim_cas/numsim_cas.h>
#include <iostream>
int main() {
  using namespace numsim::cas;
  auto [A4] = make_tensor_variable(std::tuple{"A4", std::size_t{3}, std::size_t{4}});
  auto r = inv(A4);
  std::cout << "inv(rank-4 symbol) = " << to_string(r) << " rank=" << r.get().rank() << "\n";
  // Output: inv(rank-4 symbol) = inv(A4) rank=4
  // No throw, no warning. A subsequent evaluator pass would either fail
  // at tmech::inv (which expects rank-2) or silently produce wrong results.
}
```

## The misleading comment

In the current `inv()`:

```cpp
template <tensor_expr_holder Expr>
[[nodiscard]] constexpr inline auto inv(Expr &&expr) {
  if (is_same<tensor_inv>(expr))
    return expr.template get<tensor_inv>().expr();
  if (is_same<identity_tensor>(expr))
    return std::forward<Expr>(expr);
  // …
  // Rank-2 only; rank-4 inverses are a separate concern.
  if (expr.get().rank() == 2 && expr.get().dim() % 2 != 0 &&
      contains_skew_factor(expr)) {
    throw invalid_expression_error(
        "inv: operand contains a skew-symmetric factor in odd dimensions "
        "(singular)");
  }
  return make_expression<tensor_inv>(std::forward<Expr>(expr));
}
```

The "Rank-2 only" comment is attached to the skew-symmetric singularity guard, **but the final `make_expression<tensor_inv>` line has no rank guard at all** — non-rank-2 inputs sail through and build a symbolic inv node. The comment misleads readers into thinking the function rejects non-rank-2 input.

## Why this matters

1. **Silent misbehaviour at evaluation.** `tensor_inv` evaluation almost certainly relies on `tmech::inv`, which is rank-2-only. A rank-4 `tensor_inv` will either throw a less-helpful runtime error from tmech or produce undefined results.

2. **Mathematical ambiguity.** "Inverse" of a rank-4 tensor isn't well-defined without a fixed contraction convention. The minor-identity convention (used elsewhere in this codebase, e.g. for `identity_tensor` at rank 4) is one valid choice, but the simplifier doesn't compute it — it just stores the symbolic node.

3. **Spread to derived expressions.** `inv(A4)` flowing into a larger expression bakes the bad state in and surfaces it later, far from the actual construction site.

4. **Surfaces during PR review.** Found during the third-pass review of #189 (kronecker_delta → identity_tensor unification). The identity_tensor short-circuit at rank-4 (`inv(I4) → I4`) is fine because it returns the operand unchanged; the surprise is that a rank-4 *non-identity* symbol gets through.

## Suggested fix

Add a rank check at the top of `inv()`, parallel to the existing skew-symmetric guard:

```cpp
if (expr.get().rank() != 2) {
  throw invalid_expression_error(
      "inv: only rank-2 tensors are supported (got rank " +
      std::to_string(expr.get().rank()) + ")");
}
```

Place it *after* the `tensor_inv` short-circuit (so `inv(inv(A))` still collapses) and *after* the `identity_tensor` self-inverse short-circuit (so `inv(I4)` still works — it's mathematically meaningful for the minor identity, even though general rank-4 inversion isn't).

Update the misleading comment on the skew guard from "Rank-2 only; rank-4 inverses are a separate concern" to something like "The skew guard is the rank-2 singularity check; rank-4 inputs are already rejected above."

## Acceptance criteria

- [ ] `inv()` throws `invalid_expression_error` for rank != 2 inputs, *except* for `identity_tensor` (any even rank, self-inverse) and `tensor_inv` (which collapses to the inner).
- [ ] Lock-in tests:
  - `inv(rank-4 symbol)` throws with a message containing `"rank"`.
  - `inv(make_expression<identity_tensor>(3, 4))` still returns the operand unchanged (rank-4 minor-identity self-inverse).
  - `inv(inv(rank-4 symbol))` — what's the right behaviour? Probably: the inner `inv(rank-4 symbol)` throws at construction, so the outer never sees it. Lock that in.
- [ ] Updated comment on the skew-symmetric guard removing the misleading "rank-4 inverses are a separate concern" line.
- [ ] CHANGELOG entry under `### Fixed`.

## Related

- Surfaced during #189 review (3rd pass).
- See also #187 (`inv(tensor_zero)` should throw) — same kind of fix, parallel singularity guard. These two could be bundled into one "harden `inv()` construction-time guards" PR.

petlenz:
Re-opened after revert PR #199.

Originally landed as #193 (squash `4d997c2`) on 2026-05-25; reverted as part of #199 because user wanted per-PR review rather than batch merge.

## Note on stacking

This branch is the bottom of a 7-PR stack reopened from the reverted batch:

```
main ← 193-reopened ← 189-reopened ← 186-reopened ← 194-reopened ← 195-reopened ← 196-reopened ← 197-reopened
```

Merge **#199 (the revert PR) first**, then this PR. After this PR merges with `--delete-branch`, GitHub will auto-retarget #189-reopened to `main`.

## Original content

Two construction-time guards on `inv()`:
1. `inv(tensor_zero)` → throws `"singular"` (closes #187)
2. `inv(rank != 2)` → throws `"only rank-2 tensors are supported"` (closes #192)

See the original PR #193 description for full details and test plan.

Closes #187
Closes #192


petlenz:
Closing — the original work this PR was reopening is still on main (#199 was never merged, so the revert never happened). Re-applying would be a no-op or conflict against the existing commits. Confirmed via `git log origin/main` on 2026-05-29.

PR #204, opened 2026-05-26
Re-open: Fix diff(tanh(x), x) throwing duplicate-child error (closes #180)

### LINKED ISSUE #180: diff(tanh(x), x) throws "duplicate child insertion" in n_ary_tree
## Bug

```cpp
auto [x] = make_scalar_variable("x");
auto d = diff(tanh(x), x);  // throws
// what():  n_ary_tree::insert_hash: duplicate child insertion
```

Discovered while writing a lock-in test for #33's composed hyperbolic functions. The structural form of `tanh(x)` is currently

```
2*(exp(x)-exp(-x))/(2*(exp(x)+exp(-x)))
```

— sinh(x) over cosh(x), both sharing `exp(x)` and `exp(-x)` subexpressions. Differentiating via the quotient rule produces add nodes that contain the same child key twice (e.g. two `exp(x)` entries appearing through different paths), tripping `n_ary_tree::insert_hash`'s duplicate guard.

## Why this matters

#33 added `tanh` (along with the other hyperbolics) as a composition of primitives, expecting "differentiation derives automatically." That promise holds for `sinh` and `cosh` individually but breaks for `tanh` because of the shared-subexpression structure in the composition.

```cpp
diff(sinh(x), x)  // works, returns cosh-equivalent
diff(cosh(x), x)  // works, returns sinh-equivalent
diff(tanh(x), x)  // throws
```

## Possible fixes

1. **Fix the diff path** so it handles shared sub-expressions without producing duplicate-key adds. May require teaching the chain rule / quotient rule to use `merge_or_insert` consistently.
2. **Implement tanh as exp-based** with structural canonicalisation that avoids the shared subexpression (e.g. `tanh(x) = (exp(2x) - 1) / (exp(2x) + 1)`). Diff might then succeed.
3. **Give `tanh` a dedicated AST node** with an explicit derivative — bypasses the composition issue at the cost of one new node type.

Option 2 is the least invasive; option 1 is the most general.

## Acceptance

- [ ] `diff(tanh(x), x)` does not throw.
- [ ] Result evaluates numerically to `1 / cosh²(x)` (= sech²(x)) within tolerance.
- [ ] Lock-in test added: see the partial `ScalarHyperbolicDerivativesMatchClosedForm` in CoreBugFixTest.h which currently tests only sinh because of this bug.

Surfaced during cluster-3 critical review of PR #179.

petlenz:
Re-opened after revert PR #199. Stacked on #194-reopened.

Originally landed as #195 (squash `faff15c`) on 2026-05-25; reverted as part of #199.

See the original PR #195 description for full details.

Closes #180

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ and from v0.1.0 onward the project adheres to [Semantic Versioning](https://semv
 - Replaced the regression test for #142 with one that actually exercises int64 overflow (the original test's numerators were divisible by 3 and normalized away from the overflow path). Closes #170.
 - `inv()` now rejects the zero tensor at construction (`inv(tensor_zero)` and the composite `inv(0 · A)` form) with a clear "singular" error instead of silently building a symbolic `inv(0)` node that would NaN/Inf during evaluation. Closes #187.
 - `inv()` now rejects rank > 2 inputs at construction. Previously the function accepted them and built a symbolic `tensor_inv` node that would fail at the tmech evaluator (which is rank-2 only) or silently produce wrong results. The `identity_tensor` short-circuit still fires first, so the rank-4 minor identity remains self-inverse. Closes #192.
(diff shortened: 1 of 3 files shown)


petlenz:
Closing — the original work this PR was reopening is still on main (#199 was never merged, so the revert never happened). Re-applying would be a no-op or conflict against the existing commits. Confirmed via `git log origin/main` on 2026-05-29.

Issue #139, opened 2026-05-17
[umbrella] Constitutive modeling primitives needed for downstream code generation


petlenz:
## Background

A sister repo (working name **numsim-cas-codegen**) will be created to generate MOOSE-compatible constitutive Material classes from numsim-cas symbolic expressions. Implementation language: C++ throughout (no Python recipe layer). The codegen will:

1. Accept a symbolic recipe declared via a C++ class (registry for inputs, parameters, history, internals, outputs).
2. Differentiate symbolically (using numsim-cas's existing differentiation) to obtain consistent tangents.
3. Run pointer-based CSE over the resulting DAG.
4. Emit a tmech-typed C++ function body.
5. Wrap that body in MOOSE Material boilerplate with proper boundary conversions (\`RankTwoTensor\` ↔ \`tmech::tensor<Real, D, R>\` at the boundary; tmech internally).

## What's needed from numsim-cas

This umbrella tracks the symbolic-side primitives that don't exist yet. Each is a separate issue:

### Required for any branching constitutive logic

- **#136** — Scalar comparison operators and boolean primitives (predicate domain). Foundational; should land first.
- **#135** — \`if_then_else\` ternary node across scalar / tensor / t2s domains. Depends on #136.

### Required for damage, contact, ramp-style models

- **#137** — \`min\` and \`max\` nodes for scalar domain. Depends on #135 for the differentiation rule.
- **#138** — Macauley bracket and ramp / smoothing functions (composed from #135 + #137).

## What is NOT needed from numsim-cas

These belong in the codegen repo, not here:

- The code-emitter visitor itself (a new \`numsim::cas::visitor_const\`-derived class in the codegen repo).
- The CSE pass over the DAG (codegen-side).
- The Recipe / ConstitutiveModel C++ class with the registry for inputs / outputs / history / parameters / internals.
- MOOSE Material text templating.
- The Newton-loop generation for return-mapping plasticity (built on top of the recipe + symbolic differentiation already in numsim-cas).
- Boundary conversion helpers (\`to_tmech(RankTwoTensor)\` / \`from_tmech\`).
- Build-system integration with MOOSE apps.

## Phasing

| Phase | numsim-cas work | codegen-repo work |
|-------|-----------------|-------------------|
| A | — | Recipe class with input/parameter/output. Codegen visitor. CSE. Hyperelasticity end-to-end. |
| B | — | Add history (old/new) to recipe. Viscoelasticity end-to-end. |
[...]

Issue #135, opened 2026-05-17
Add if_then_else ternary node across scalar, tensor, and tensor_to_scalar domains


petlenz:
**Revised plan (Option B).** Originally scoped against a dedicated predicate domain — see PR #140 (closed) for that variant. Switching to scalar/t2s-valued conditions that evaluate to 0.0 or 1.0.

## Required scope

New AST node `scalar_if_then_else(cond, then, else)` and `tensor_to_scalar_if_then_else(cond, then, else)` and `tensor_if_then_else(cond, then, else)` — one per host domain.

The **condition** is in the same domain as the branches for scalar / t2s (since predicates ARE scalars/t2s in Option B). For tensor, the condition is a scalar (or t2s acting as scalar) — there's no notion of tensor-valued boolean.

| Node type | Condition | Then / Else |
|---|---|---|
| `scalar_if_then_else` | scalar | scalar |
| `tensor_to_scalar_if_then_else` | t2s | t2s |
| `tensor_if_then_else` | scalar | tensor |

Tensor with a t2s condition is reachable via wrapping the condition through `tensor_to_scalar_scalar_wrapper` first, or by adding a fourth overload if it turns out to be common.

## Visitor coverage

Each new node must be added to:
- `NUMSIM_CAS_*_NODE_LIST` for its domain
- Printer, latex printer, evaluator, differentiation
- The static_assert fallback in each visitor catches missing overloads at compile time

## Construction-time simplification

- `if_then_else(scalar_zero, a, b) → b`
- `if_then_else(scalar_one, a, b) → a`
- `if_then_else(scalar_constant{c}, a, b) → a if c != 0 else b`
- `if_then_else(cond, a, a) → a`

## Differentiation

Assumes the condition does not depend on the differentiation variable:
```
d/dx if_then_else(cond, a(x), b(x)) = if_then_else(cond, da/dx, db/dx)
```

For conditions that do depend on x — strictly the derivative has Dirac contributions at the boundary, but the constitutive-model use case (yield function, contact gap) is exactly the case where the boundary is measure-zero in time evolution, so the rule above is correct in practice. Document the assumption.

## Acceptance criteria

- 3 new node types registered (one per host domain).
- Free functions `if_then_else(cond, a, b)` in each domain's `*_std.h` overloaded on the input types.
- All construction-time simplifications fire.
- Differentiation produces the expected piecewise result.
- 15+ lock-in tests covering all three domains, both arms, and the differentiation rule.

[...]

Issue #138, opened 2026-05-17
Add Macauley bracket and ramp / smoothing functions for constitutive models


petlenz:
## Motivation

Damage and contact constitutive models routinely use:

- **Macauley bracket** \`<x>+ = max(x, 0)\` — positive part. Standard in damage activation, ramp-style yield, contact pressure.
- **Negative part** \`<x>- = max(-x, 0) = -min(x, 0)\` — symmetric, e.g. plastic dissipation rate.
- **Smoothed Heaviside** for regularised plasticity (avoiding non-smooth differentiation).
- **Smoothed Macauley** \`(x + sqrt(x² + ε²))/2\` — \`C^∞\` regularisation of the positive part.

These compose from existing / planned primitives so they don't require new AST node types — they go in \`scalar_std.h\` as free functions, similar to the hyperbolic functions added in PR #123.

## Required scope

Free functions in \`scalar_std.h\`:

\`\`\`cpp
template <scalar_expr_holder E> auto macauley_plus(E const &e);  // max(e, 0)
template <scalar_expr_holder E> auto macauley_minus(E const &e); // -min(e, 0)
template <scalar_expr_holder E> auto heaviside(E const &e);      // 0 if e<0 else 1, via if_then_else
template <scalar_expr_holder E, scalar_expr_holder Eps>
auto smoothed_macauley(E const &e, Eps const &eps);  // (e + sqrt(e² + eps²)) / 2
\`\`\`

Plus a tensor-domain variant operating componentwise where it makes sense — though for our use cases the operand is typically already a scalar (yield function value, contact gap), so scalar versions suffice for MVP.

## Construction-time simplifications

- \`macauley_plus(constant)\` → numeric value
- \`macauley_plus(macauley_plus(x))\` → \`macauley_plus(x)\` (idempotent)
- \`macauley_plus(-x)\` → \`macauley_minus(x)\`

## Acceptance criteria

- Four new free functions defined as compositions.
- Lock-in tests verifying numeric evaluation at representative points.
- Codegen renders them sensibly (composed form is fine — no need for dedicated nodes).

## Depends on

- #137 (min/max nodes) for the natural implementation of \`macauley_plus\` and \`macauley_minus\`.
- #135 (if_then_else) for \`heaviside\`.

## Out of scope

- Dedicated node types — these are composition-only, mirroring PR #123's hyperbolic approach. If named-form printing becomes important later, file as a follow-up like #126.

petlenz:
[...]

PR #207, opened 2026-05-26
Add scalar max / min AST nodes (closes #137)

### LINKED ISSUE #137: Add min and max nodes for the scalar domain
## Motivation

Standard CAS primitives. Required for several constitutive forms:
- **Macauley bracket** \`<x>+ = max(x, 0)\` — damage models, ramp-style yield surfaces, regularised penalty contact.
- **Clamping** \`max(lo, min(hi, x))\` — bounded internal variables (saturation hardening, damage variable cap at 1).
- **L-infinity-style norms** — \`max(|σ_1|, |σ_2|, |σ_3|)\` for stress measures.

Could be derived from \`if_then_else\` (\`max(a, b) = if(a > b, a, b)\`) but having dedicated nodes is cleaner:
- Better printed form (\`max(a, b)\` vs \`a > b ? a : b\`).
- Pattern-matchable in simplifier (e.g. \`max(0, max(x, 0)) → max(x, 0)\`).
- Standard codegen target (\`std::max(a, b)\` in generated C++).

## Required scope

**New nodes:** \`scalar_max\`, \`scalar_min\` in the scalar node list.

**Visitor coverage:** printer, latex_printer, evaluator, differentiation (the static_assert fallback will force this).

**Operators:** \`max(a, b)\`, \`min(a, b)\` free functions in \`scalar_std.h\`.

**Construction-time simplification:**
- \`max(x, x) → x\`
- \`min(x, x) → x\`
- \`max(c1, c2) → constant\` (numeric folding)
- \`min(c1, c2) → constant\`
- \`max(x, -inf) → x\`, \`min(x, +inf) → x\` (when bounds appear as constants)

**Differentiation:** piecewise, using \`if_then_else\` once #135 lands:
\`\`\`
d/dx max(a(x), b(x)) = if_then_else(a > b, da/dx, db/dx)
d/dx min(a(x), b(x)) = if_then_else(a < b, da/dx, db/dx)
\`\`\`
Until #135 lands, the differentiation rule can throw \`not_implemented_error\` or fall back to the sub-gradient at the \`a == b\` boundary.

**Evaluator:** \`std::max\` / \`std::min\` direct dispatch.

## Acceptance criteria

- Two new node types registered.
- All scalar visitors have explicit overrides.
- Construction-time simplifications fire.
- Lock-in tests for: numerical evaluation, simplifications, codegen rendering as \`std::max(...)\`.

## Depends on

- #135 (if_then_else) for the differentiation rule. Can land without #135 and have differentiation throw until #135 is in.

petlenz:
## Summary

Closes #137. Two new binary scalar nodes — `scalar_max` and `scalar_min` — modeling `std::max(a, b)` / `std::min(a, b)`. Required by the constitutive-modelling cluster (#139):

- **Macauley positive part:** `<x>+ = max(x, 0)` — damage activation, ramp yield, contact pressure.
- **Clamping:** `max(lo, min(hi, x))` — bounded internal variables.
- **L-infinity norms:** `max(|σ_1|, |σ_2|, |σ_3|)` — stress measures.

## Implementation

| File | Change |
|------|--------|
| `include/numsim_cas/scalar/scalar_max.h` (new) | `class scalar_max final : binary_op<...>` |
| `include/numsim_cas/scalar/scalar_min.h` (new) | `class scalar_min final : binary_op<...>` |
| `scalar_node_list.h` | Added `NEXT(scalar_max)`, `NEXT(scalar_min)` |
| `scalar_definitions.h` | Includes added |
| `scalar_std.h` | `max(a, b)` / `min(a, b)` free functions with construction-time folds |
| `scalar_evaluator.h` | `std::max` / `std::min` direct dispatch |
| `scalar_printer.{h,cpp}` | `"max(a, b)"` function-call form |
| `scalar_latex_printer.{h,cpp}` | `\\max\\!\\left( a, b \\right)` typeset form |
| `scalar_rebuild_visitor.h` | Rebuild via `max()` / `min()` to re-trigger simplification |
| `scalar_differentiation.h` | Throws `not_implemented_error` until #135 lands |
| `scalar_assumption_propagator.{h,cpp}` | Recursive + shallow inference (real iff both operands real) |
| `scalar_limit_visitor.{h,cpp}` | Reports `unknown` |
| `core/contains_expression.cpp` | Recurse into both children |
| `tests/ScalarExpressionTest.h` | 10 new lock-in tests |

## Construction-time simplifications

```cpp
max(x, x)     → x              // idempotent on hash-equal operands
max(c1, c2)   → numeric        // constant folding via numeric_less
max(x, y)     == max(y, x)     // commutative canonical form (operand sort by hash)
```

The commutative canonical form was added during self-review when I spotted `max(x, y).hash != max(y, x).hash`. Without it, downstream rules like `max(a, b) + max(b, a) → 2*max(a, b)` would never fire.

## Differentiation

Pre-#135 the piecewise derivative `d/dx max(a, b) = if_then_else(a > b, da/dx, db/dx)` can't be expressed symbolically, so the visitor throws `not_implemented_error`. The `MaxDiffThrowsUntilIfThenElseLands` test pins this contract so when #135 lands, the upgrade is forced and visible.

## Known limitation

`max(max(x, y), z)` and `max(x, max(y, z))` produce different structural forms — true associative flattening would require converting max/min to n_ary nodes (like `scalar_add` / `scalar_mul`), out of scope here. The issue's acceptance criteria don't require it.

## Tests

10 new lock-in tests in `ScalarExpressionTest.h`:

- `MaxBasicPrint`, `MinBasicPrint` — print form
- `MaxIdempotentOnEqualOperands`, `MinIdempotentOnEqualOperands` — `max(x, x) → x` including compound operands
- `MaxConstantFolding`, `MinConstantFolding` — numeric folding for both signs
- `MaxMinEvaluator` — round-trip via `scalar_evaluator<double>`
- `MaxMacauleyPlusEvaluator` — the motivating constitutive-modelling case
- `MaxMinCommutativeCanonicalForm` — `max(x, y) == max(y, x)` including compound operands
- `MaxDiffThrowsUntilIfThenElseLands` — pins the pre-#135 throw contract

## Test plan

- [x] Clean build under `-Werror`.
- [x] `ctest --test-dir build` — **1747/1747 passing** (was 1737 + 10 new).
- [x] Edge-case probes (see review notes in PR description): substitution, contains_expression, mixed numeric/symbolic ordering, evaluator on chains.

Closes #137


petlenz (commented):
## Critical review — found 2 real gaps + 1 doc concern

### 1. [Real, fix on this PR] Assumption propagation drops too much sign info

`scalar_assumption_propagator::operator()(scalar_max)` and `(scalar_min)` both write `m_result = numeric_assumption_manager{}` after visiting children. The `shallow_inference_visitor::handle_minmax` is slightly better — it propagates `real_tag{}` when both children are real — but neither captures the well-known sign-monotonicity facts:

- `max(a, b) ≥ max(a, b's lower bound)`, so:
  - both `nonnegative` → `max` is `nonnegative`; both `positive` → `max` is `positive`
  - either `nonnegative` → `max` is `nonnegative` (max is dominated by the larger; if any side is ≥ 0, max ≥ 0)
  - either `positive` → `max` is `positive`
  - both `nonpositive` → `max` is `nonpositive`; both `negative` → `max` is `negative`
- Symmetric for `min`:
  - both `nonpositive` → `min` is `nonpositive`; both `negative` → `min` is `negative`
  - either `nonpositive` → `min` is `nonpositive`
  - either `negative` → `min` is `negative`
  - both `nonnegative` → `min` is `nonnegative`; both `positive` → `min` is `positive`

This matters because the Macauley bracket `<x>+ = max(x, 0)` from #138 is provably `nonnegative` and the downstream simplifier should be able to use that — currently it can't because the assumption is dropped. Concrete victim: `sqrt(max(x, 0))` won't trigger any `sqrt` real-domain rules because the argument is treated as unknown sign.

### 2. [Doc] No mention of ADL clash with `std::max` / `std::min`

The tests use `using numsim::cas::max;` inside each `TEST_F` body to disambiguate from `std::max` — but the node's doxygen doesn't tell a user this is necessary. Worth one sentence in `scalar_max.h` / `scalar_min.h` so the first user who hits "ambiguous overload" knows why.

### 3. [Noted, will fix in #209] `MaxDiffThrowsUntilIfThenElseLands` is a deliberate landmine

By design — the throw test pins the contract so #135 (PR #209) updating the diff rule is forced visible. PR #209 already renames it to `MaxDiffReturnsIfThenElsePiecewise`. No action on this PR.

### Out of scope

- `scalar_max` / `scalar_min` are `binary_op`, not `n_ary` — `max(max(a, b), c)` and `max(a, max(b, c))` stay structurally distinct. Possibly worth an n-ary upgrade later (would compose with the canonical-sort already in place) but not in this PR.
- NaN behaviour of `std::max(NaN, x)` in the evaluator is implementation-defined. Not part of the contract; live with it.

Fixing item 1 + item 2 next.

petlenz:
## Critical review

Solid work. Visitor coverage is complete, the canonical-form sort is the right call, and the lock-in tests cover idempotence, commutativity, constant folding, evaluator round-trip, and the motivating Macauley case. Three findings:

### Real

**F1 (P3) — `scalar_max` doxygen says "Differentiation uses the sub-gradient at the boundary `a == b`"** but the implementation throws `not_implemented_error` until #135 lands. Sub-gradient language is appropriate AT the boundary but misleads on what the code actually does. Suggest tightening to: "Differentiation requires `if_then_else` (#135); throws until that lands; once available, returns `if_then_else(a > b, da/dx, db/dx)`."

**F2 (P3) — `MaxBasicPrint` expects `"max(x,y)"` but the canonical sort orders operands by hash.** If `hash(y) < hash(x)` (e.g. after a libstdc++ change to `std::hash<string>` seeding, or in any env where it's randomized), the test prints `"max(y,x)"` and fails. Same applies to `MinBasicPrint`. Either document the hash-order dependence in the test comment, or accept either order via `EXPECT_THAT(..., AnyOf("max(x,y)", "max(y,x)"))`.

**F3 (P4) — `not_implemented_error` for max/min diff is a great "force the upgrade" signal**, but for users hitting it before #135 lands, a workaround hint in the message (`smoothed_macauley(e, ε)`, or hand-rewrite via if_then_else once available) would shorten the debugging loop. Pure-doc fix.

### Minor / accepted

- The ADL-clash doc on `max` is excellent — same caveat applies to `min`'s body but the cross-reference handles it.
- Sign-monotonicity propagator helpers are correctly shared across the recursive + shallow visitors via the anonymous namespace — DRY without leaking symbols.
- `MaxIdempotentOnEqualOperands` correctly tests compound operands too (`max(x+y, x+y)`), not just literal-AST-equal cases.

LGTM modulo F1/F2.

Issue #136, opened 2026-05-17
Add scalar and tensor_to_scalar comparison operators (returning 0/1)


petlenz:
**Revised plan (Option B from design discussion).** Originally scoped as a dedicated predicate domain — see PR #140 (closed) for that variant. Switching to scalar/t2s-valued predicates that evaluate to 0.0 or 1.0.

## Motivation

\`if_then_else\` (#135) needs a condition node. Constitutive models also benefit from indicator-function patterns like \`(yield_function > 0) * plastic_stress + (yield_function <= 0) * elastic_stress\` which require comparisons as scalar expressions.

## Why Option B (scalar predicates) over a dedicated domain

- **~3× less infrastructure.** No fourth domain, no separate visitor base, no cross-cutting impact on existing visitors (codegen, simplifier audits, LeafCollector).
- **Indicator-function modelling works directly.** \`(a < b) * x\` is a legitimate symbolic expression — useful for damage activation, smoothed yield, penalty contact.
- **Differentiation is implicit.** Comparisons differentiate to 0 almost everywhere (sub-gradient at the boundary). Existing scalar differentiation rules handle this without special-casing.
- **Codegen quality unchanged.** The codegen visitor inspects the node type and emits \`(a < b)\` as a C++ bool, not the runtime 0/1 representation.

## Scope

Add 12 new node types — 6 in scalar domain, 6 in t2s domain. Each is a \`binary_op\` with both operands in the host domain, evaluating to 0.0 or 1.0.

### Scalar comparisons

- \`scalar_lt(a, b)\` — \`a < b\`
- \`scalar_gt(a, b)\` — \`a > b\`
- \`scalar_eq(a, b)\` — \`a == b\` (numerical equality; symbolic equality is \`expression_holder::operator==\`)
- \`scalar_le(a, b)\` — \`a <= b\`
- \`scalar_ge(a, b)\` — \`a >= b\`
- \`scalar_ne(a, b)\` — \`a != b\`

Free-function API (no operator overload, to avoid disrupting \`expression_holder\`'s hash-based \`<\`):

\`\`\`cpp
expression_holder<scalar_expression> lt(scalar_expr_holder auto&& a, scalar_expr_holder auto&& b);
// gt, eq, le, ge, ne similarly
\`\`\`

### T2s comparisons

Mirror of scalar — six nodes \`tensor_to_scalar_lt\` etc. taking two t2s operands. Free functions overloaded on the t2s domain.

[...]

PR #208, opened 2026-05-26
Add Macauley bracket / ramp / Heaviside / smoothed Macauley (closes #138)

### LINKED ISSUE #138: Add Macauley bracket and ramp / smoothing functions for constitutive models

petlenz:
## Summary

Closes #138. Four constitutive-modelling primitives composed from existing nodes — no new AST types, mirroring PR #123's hyperbolic-functions approach.

**Stacked on #207 (#137 min/max nodes).** Merge #207 first; this PR's base will auto-retarget to `main`.

## API

```cpp
auto macauley_plus(E const &e);   // <e>+ = max(e, 0)
auto macauley_minus(E const &e);  // <e>- = -min(e, 0)
auto heaviside(E const &e);       // 0 if e<0, 1 otherwise (right-continuous step)
auto smoothed_macauley(E const &e, Eps const &eps);  // (e + sqrt(e² + ε²)) / 2
```

## Implementation choices

**`heaviside` via `ge(e, 0)`, not `if_then_else`.** The issue body suggested the `if_then_else` route (deferred to #135), but the comparison node from #136 produces the same `0`/`1` indicator with no #135 dependency. Once #135 lands, both routes collapse to equivalent comparison combinations.

**`smoothed_macauley` is fully C^∞.** As `ε → 0` recovers `max(e, 0)`. For nonzero `ε`, the second-order derivative exists everywhere — useful for Newton-style solvers that choke on the kink at `e = 0` in plasticity / damage models.

## Construction-time folds

Per #138 acceptance criteria:

- `macauley_plus(-x) → macauley_minus(x)` (via `scalar_negative` detection)
- `macauley_minus(-x) → macauley_plus(x)` (symmetric)
- `macauley_plus(macauley_plus(x)) → macauley_plus(x)` (idempotence — the positive part is non-negative by construction)
- `macauley_plus(constant)` collapses via the underlying `max()` folding

The idempotence rule was added during self-review when probing surfaced `macauley_plus(macauley_plus(x))` producing `max(0, max(0, x))` instead of folding.

## Tests

10 new lock-in tests in `ScalarExpressionTest.h`:

| Test | Verifies |
|------|----------|
| `MacauleyPlusEvaluatorPositive/Negative/AtZero` | Numerical evaluation at `+3`, `−2`, `0` |
| `MacauleyMinusEvaluatorPositive/Negative` | `<3>− = 0`, `<−2>− = 2` |
| `MacauleyPlusIdempotent` | `<<x>+>+ = <x>+` |
| `MacauleyPlusOfNegatedFoldsToMacauleyMinus` | Symmetric folds |
| `HeavisideEvaluatesAsRightContinuousStep` | `H(−0.1) = 0`, `H(0) = 1`, `H(1.5) = 1` |
| `SmoothedMacauleyConvergesToMacauleyPlusAsEpsShrinks` | Monotone convergence at `x = −1` |
| `SmoothedMacauleyAtPositive` | `≈ 1` at `x = 1`, small `ε` |

## Test plan

- [x] Clean build under `-Werror`
- [x] `ctest --test-dir build` — **1759/1759 passing** (was 1749 + 10 new)
- [x] Numerical probes verify the convergence direction and Heaviside step behaviour

Closes #138


petlenz (commented):
## Critical review — 1 doc accuracy issue, 2 coverage gaps, 1 asymmetry to document

### 1. [Doc] `heaviside` docstring overclaims

The header for `heaviside` says:
> Once #135 lands, both routes simplify to indicator + comparison combinations that the simplifier can collapse uniformly.

This is hopeful — but the comparison-indicator path (`ge(e, 0)`) doesn't actually share simplifier coverage with the `if_then_else` form yet. There's no merged simplifier rule that collapses them. Either land that rule or soften the claim. Action: soften the wording so it doesn't promise behaviour we haven't built.

### 2. [Asymmetry to document] `macauley_minus(macauley_minus(x))` does NOT fold to `macauley_minus(x)`

The positive bracket has an idempotence guard (`<<x>+>+ → <x>+`), but the negative bracket has none — by symmetry the user might expect `<<x>->- → <x>-`. But the mathematical answer is different:

```
<x>-  = max(0, -x)  ≥ 0  for all x
<<x>->- = max(0, -<x>-) = max(0, -nonneg) = max(0, nonpos) = 0
```

So `<<x>->- = 0`, not `<x>-`. The current implementation evaluates to 0 correctly but doesn't fold symbolically. This is the *mathematically correct* asymmetry, not a bug — but the doxygen doesn't mention it and a user reading `macauley_plus` then `macauley_minus` will rightly wonder why one has an idempotence guard and the other doesn't. Action: add a note to `macauley_minus`'s docstring explaining the asymmetry.

### 3. [Coverage gap] No assumption-propagation tests for the Macauley brackets

Once #207's fix flows through (sign-monotonicity propagation across `scalar_max` / `scalar_min`), `macauley_plus(x)` is provably nonnegative — but no test pins this. Same for `macauley_minus(x)`. This is the cross-PR validation that the #207 fix actually pays off downstream. Action: add two tests in the assumption fixture.

### 4. [Pre-existing, will refresh after rebase] `macauley_plus` doxygen mentions throwing

The doc says "Differentiation goes through the underlying `max()` and throws until `if_then_else` (#135) lands". After #135 lands (PR #209), this needs updating. Will refresh on rebase.

### Out of scope

- Folding `min(nonneg, 0) → 0` would let the symbolic form of `<<x>->-` collapse to 0 — but requires assumption-aware folding at construction time, which is a substantial follow-up.

Working off items 1–3 now.

petlenz:
## Critical review

Composition-only approach mirrors #123 cleanly — no new AST means no new visitor coverage to maintain, and the idempotence-asymmetry doxygen on `macauley_minus` is excellent for future maintainers. Two findings:

### Real

**F1 (P3) — `macauley_plus`'s `<<x>+>+ → <x>+` fold matches any `max(·, 0)` or `max(0, ·)`, not just Macauley brackets.** A user-built `max(some_known_negative_quantity, 0)` is also detected and treated as "already bracketed". Mathematically still correct (a max with 0 IS nonneg), but the comment "Detect the inner macauley shape (a max with one operand being 0)" understates how broadly this fires. Worth noting that the fold's trigger is structural, not semantic.

**F2 (P4) — `smoothed_macauley(e, 0)` doesn't degenerate to `max(e, 0)` automatically.** It computes `(e + sqrt(e²)) / 2 = (e + |e|) / 2`, mathematically equal to `max(e, 0)`, but `sqrt(e²)` doesn't fold to `abs(e)` and there's no `(e + abs(e))/2 → max(e, 0)` rule. A user passing `eps=0` (perhaps from a parametric sweep that shrinks ε to zero) gets a more complex symbolic form than they'd expect. Cheap fix: detect `is_same<scalar_zero>(eps)` at the top and delegate to `macauley_plus(e)`.

### Minor / accepted

- Heaviside-via-ge is the right call. Avoiding the #135 dep here means this PR isn't blocked on #209 either.
- The convergence test asserting monotone decrease in ε is the right shape — much stronger than a single-point limit check.
- Sign assumptions correctly flow through to `macauley_plus` / `macauley_minus` via #207's helpers; the cross-validating tests are well-placed.

LGTM modulo F2 (cheap fix; F1 is a comment tightening).

PR #209, opened 2026-05-26
Add scalar if_then_else (partial #135 — scalar half)

petlenz:
## Summary

Adds the scalar-domain piece of #135's if_then_else family. Stacked on **#208** (#138 Macauley).

This PR delivers `scalar_if_then_else` + visitor coverage + `if_then_else()` free function in `scalar_std.h`. The `tensor_to_scalar_if_then_else` and `tensor_if_then_else` variants listed in #135's acceptance criteria are **deferred to a follow-up issue** (filed alongside this PR) — same pattern as our earlier scoping decisions where each domain warrants its own review cycle.

## Design choice — "Option B" predicates

The condition is a scalar evaluating to `0.0` for false and any non-zero value (canonically `1.0`) for true — matching the indicator convention of the comparison nodes from #136. No separate predicate domain; the condition lives in the same domain as the branches.

## Implementation

| New / Touched | Purpose |
|---|---|
| `include/numsim_cas/core/ternary_op.h` (new) | Generic three-operand base; condition base can differ from branch base (prepares tensor_if_then_else where cond is scalar, branches are tensor) |
| `include/numsim_cas/scalar/scalar_if_then_else.h` (new) | The scalar node |
| `scalar_node_list.h`, `scalar_definitions.h`, `scalar_std.h` | Registration + free function |
| 8 visitor sites | evaluator (lazy on unselected), printer, latex_printer, rebuild, diff, assumption_propagator (both classes), limit_visitor, contains_expression |
| `scalar_differentiation.cpp` | New rules including the **max/min diff upgrade** (closes the gap from #137) |

## Construction-time folds

```cpp
if_then_else(scalar_zero, a, b)        → b
if_then_else(scalar_one, a, b)         → a
if_then_else(scalar_constant{c≠0}, a, b) → a
if_then_else(cond, a, a)               → a   // branches identical
```

## Max / min diff upgrade

PR #207's `MaxDiffThrowsUntilIfThenElseLands` test threw `not_implemented_error` pending this PR. Now:

```cpp
d/dx max(a, b) = if_then_else(a > b, da/dx, db/dx)
d/dx min(a, b) = if_then_else(a < b, da/dx, db/dx)
```

Test renamed `MaxDiffReturnsIfThenElsePiecewise` and rewritten to assert the new structural shape + evaluator round-trip.

## Lazy evaluation

The evaluator only applies the selected branch — load-bearing for damage / contact models where the unselected arm may contain expressions that would error at evaluation (e.g. `log(x)` for `x ≤ 0`). Locked in by `IfThenElseEvaluatorLazyOnUnselectedBranch`.

## Tests

7 new lock-in tests in `ScalarExpressionTest.h` + the upgraded `MaxDiffReturnsIfThenElsePiecewise`. Full suite: **1766/1766 passing** (was 1759 + 7 new).

## Not yet in this PR

- `tensor_to_scalar_if_then_else` — same pattern, t2s domain.
- `tensor_if_then_else` — branches are tensor, condition is scalar (`ternary_op<BaseCond, BaseThen>` with different bases — the ternary_op header already supports this).

Follow-up issue will track. #135 stays open until those land.

Stacked on #208.


petlenz (commented):
## Critical review — 1 real gap (assumption prop), 1 cleanup

### 1. [Real, fix on this PR] `if_then_else` assumption propagation drops branch-intersection facts

Same shape as the #207 finding. `scalar_assumption_propagator::operator()(scalar_if_then_else)` writes `m_result = numeric_assumption_manager{}` after visiting children. The shallow_inference_visitor partially compensates by propagating `real_tag` when both branches are real, but it drops the sign-intersection facts that fall out trivially.

The result of `if_then_else(cond, A, B)` is either `A` or `B`, so the intersection of their assumption sets holds in either case:

- both branches `nonnegative` ⇒ result `nonnegative`
- both branches `positive`    ⇒ result `positive`
- both branches `nonpositive` ⇒ result `nonpositive`
- both branches `negative`    ⇒ result `negative`
- both branches `nonzero`     ⇒ result `nonzero`
- both real → real (already done in shallow)
- both integer → integer

Concrete payoff: with #207's max/min sign-propagation in place, `max(x, 0)` is nonnegative. With the corresponding rule here, the symbolic form produced by `diff(max(x, y), x)` — which is `if_then_else(gt(x,y), 1, 0)` — would be inferable as nonnegative. Currently it's not.

### 2. [Doc] `try_extract_scalar_number` recurses through `scalar_negative`

`if_then_else(scalar_constant{c}, a, b)` folds via `try_extract_scalar_number`, which transparently handles `scalar_negative(scalar_constant{c})` too. The fold docs should mention this — a user reading the doxygen might think the fold only fires on bare constants. Minor.

### 3. [Verified, no action] Diff rule for max/min via if_then_else is correct

I verified:
- `d/dx max(a, b)` should be `da/dx` when `a > b`, `db/dx` when `a < b`. The implementation emits `if_then_else(gt(a, b), da, db)` — picks `da` when `gt(a, b) = 1`, i.e. `a > b`. ✓
- Symmetric for min using `lt(a, b)`. ✓

### Out of scope

- Cond-dependent diff (when `cond` itself depends on x): currently ignored. The Dirac contribution at the boundary is dropped. Doc says this is intentional for the constitutive use case. Reasonable.
- `if_then_else(cond, then_a, else_a)` with a nested redundant if_then_else: no fold. Premature.

Fixing item 1 now; item 2 is a one-line doc tweak in the same commit.

petlenz:
## Critical review

Closes the #137 gap cleanly with the new `ternary_op` base, and the lazy-evaluation lock-in test is the load-bearing one — `if_then_else(ge(x, 0), log(x), 0)` at `x = -1` would silently NaN-pollute without it. Three findings:

### Real

**F1 (P2 — code/comment mismatch) — `scalar_latex_printer.cpp`'s inline comment for `if_then_else` says "Typeset as a Iverson-bracket-style ternary: `\left\{ then : cond, else : \text{otherwise} \right\}`"** but the code emits `\operatorname{if\_then\_else}\!\left(...\right)` — function-call form, not Iverson. The comment looks like a leftover from an earlier design the code never matched. Pick one. (Either is fine; the function-call form is consistent with the plain-text printer, so I'd just update the comment.)

**F2 (P3) — The diff rule assumes `cond` doesn't depend on the differentiation variable** and silently drops the Dirac contribution at the boundary. The doxygen acknowledges this is acceptable for yield-function / contact-gap use cases. But there's no test pinning the chosen behavior when `cond` DOES depend on `x` — e.g. `if_then_else(gt(x, 0), x, -x)` (= `abs(x)`) diffs to `if_then_else(gt(x, 0), 1, -1)` (= `sign(x)`, with the Dirac at 0 dropped). Adding a test would document the design choice as intentional rather than incidental, and pin the contract for the eventual tensor variants.

**F3 (P3) — Lazy evaluator vs. eager differentiator asymmetry isn't called out at the diff visitor site.** The evaluator's `apply` is lazy on the unselected branch (load-bearing per the `log(-1)` test); the diff visitor applies BOTH branches eagerly because it's symbolic. Both correct, but a future maintainer reading the diff visitor in isolation might assume the laziness applies. One-line comment at the diff site would prevent the confusion.

### Minor / accepted

- `ternary_op` already supports differently-typed branches via the `BaseCond` / `BaseThen` template params — clean preparation for `tensor_if_then_else` without a refactor.
- Assumption intersection between branches handles "both nonneg ⇒ result nonneg" cleanly, and `InferMaxDiffResultIsNonnegative` cross-validates the #207 max-diff and #209 if_then_else paths together.
- `MaxDiffReturnsIfThenElsePiecewise` correctly pins behavior (numerical evaluation) rather than internal structure, so it survives a future canonicalization tweak in `max()`.

LGTM modulo F1 (real bug — fix the comment).

PR #211, opened 2026-05-26
Add curated numerical-diff golden tests (#85)

### LINKED ISSUE #85: New tests: add numerical diff checks

petlenz:
## Summary

Adds a reusable harness (`tests/NumericalDiffHelpers.h`) plus 30 named curated tests (`tests/NumericalDiffTest.h`) that verify symbolic differentiation against central-difference numerical estimates.

## Why this is not duplicate of the existing fuzz tests

The repo already has `FuzzyScalarDiffTest.h` / `FuzzyTensorDiffTest.h` which do random finite-difference checks. They are excellent at finding bugs in *combinations* of smooth operators, but:

- they give no named per-operator regression coverage,
- they explicitly **skip stiff / non-smooth expressions** (`mag > 1e5` guard, see `FuzzyScalarDiffTest.h:236`),
- they intentionally avoid junctions of operators like `max`, `min`, `macauley`, `abs`, `if_then_else`.

This PR fills both gaps with deterministic, named tests that pin one expression at one point and verify the result. When a diff rule regresses, the failing test names the operator.

## Coverage

| Family | Tests |
|---|---|
| polynomial | 1 (4 sample points) |
| trig: sin, cos, tan | 3 |
| inverse trig: asin, acos, atan (5-pt near domain edge) | 3 |
| exp, log, sqrt | 3 |
| pow (integer and fractional exponent) | 2 |
| compositions: sin(x²), exp(sin(x)), log(x²+1) | 3 |
| abs (smooth interior) | 1 |
| max / min (each branch separately) | 4 |
| macauley_plus / macauley_minus (each side) | 4 |
| smoothed_macauley (smooth at x=0) | 1 |
| if_then_else (constant cond, x-dependent cond ±) | 3 |
| compositions across operator families | 1 |
| multi-variable separability | 1 |
| **Total** | **30** |

All 30 tests pass locally. Full suite: 1766 → 1796.

## Tolerance choices

- `abs_tol = 1e-6`, `rel_tol = 1e-4` — matches `FuzzyScalarDiffTest.h:240-241`.
- `h = 1e-5` for 2-point central difference — same.
- 5-point variant (`h = 1e-3`, 4th-order accurate) for stiff-but-smooth cases like `asin(0.95)` where 2-point O(h²) truncation eats the budget.

## Harness sanity

The harness's pass/fail logic was verified by hand: at `x = 0.7`, claiming `d/dx sin(x) = -cos(x)` produces |sym - num| ≈ 1.53, far above the 1e-6 acceptance band. The harness would clearly reject any incorrect derivative.

## Out of scope

- Tensor numerical-diff golden tests — `FuzzyTensorDiffTest` already uses `tmech::num_diff_central` for the random side; a curated tensor file is a sensible follow-up but not needed to close #85 (which only mentions "numerical diff checks", and the new scalar coverage is already substantial).
- Boundary tests at non-differentiable junctions (`max(x, c)` exactly at `x = c`) — out of scope: the diff rule there is symbolic-only and the numerical comparison is meaningless across a corner.

Stacked on #209 (`135-if-then-else`). Do NOT merge yet — under review.

Closes #85.


petlenz (commented):
## Critical review — 1 missing helper, 1 doc gap, 1 dead code path

### 1. [Real, fix on this PR] No multi-variable helper — two tests roll their own

The helper exposes `EXPECT_DIFF_MATCHES(expr, var, x0)` which only binds one variable. Two tests in this PR need multi-variable evaluation and end up inlining the eight-line "perturb, evaluate, compare" pattern:

- `IfThenElseSymbolicCondPreservesNode` (lines ~260-290) binds `y` while differentiating w.r.t. `x`.
- `MultiVariableSeparability` (lines ~325-348) differentiates a 2-variable expression w.r.t. each variable in turn.

Both should use a single multi-var helper. Add `check_symbolic_diff_matches_with_env(expr, env, var, x0)` that takes a `std::map<scalar_expr, double>` for extra bindings; refactor the two callers; drop the inline loops.

### 2. [Doc] Boundary-region fragility is not documented

The tests deliberately pick evaluation points away from non-smooth junctions (e.g. `max(x, _1)` is tested at x=2 and x=0.5, never at x=1±tiny). But the harness header doesn't warn about this. A future contributor adding `EXPECT_DIFF_MATCHES(max(x, _1), x, 1.000005)` would create a flaky test: at h=1e-5, the perturbation straddles the junction at x=1, the central difference produces ~0.5 (averaging both slopes), and the symbolic answer is 0 or 1 depending on which side `1.000005` is rounded to. Worth a one-paragraph caveat in the harness's top-of-file comment.

Concretely: the rule of thumb is `|x0 - junction| > h`. Add this to the docs.

### 3. [Dead code] `EXPECT_DIFF_MATCHES_5PT` is only used in 2 places

Used by `Asin` and `Acos` at x=0.95. Not a bug — useful for any stiff-but-smooth case — but worth mentioning in its doc when *not* to reach for it (it's strictly stronger than 2-point but pays in extra eval calls). Minor.

### Out of scope

- Picking `h` adaptively (e.g. as a function of `|x0|`) — would catch a couple more edge cases but adds complexity for marginal gain.
- Per-test tolerance overrides via a varargs macro form — fine to leave at the function-call level for unusual cases.

Fixing items 1 and 2 now; item 3 is one doc line.

petlenz:
Superseded by #234 — the curated numerical-diff golden tests (`NumericalDiffHelpers.h` + `NumericalDiffTest.h`) and the if_then_else assumption-propagator additions are included verbatim in that restore PR (verified byte-identical against this branch's `tests/NumericalDiff*.h`). Closing in favor of #234 to consolidate the recovery of content lost in the #208/#209/#212 cascade-merge failure.

PR #212, opened 2026-05-26
Add Levi-Civita symbol tensor leaf node (#34)

### LINKED ISSUE #34: Introduce Levi-Civita-Symbol

petlenz:
## Summary

Adds the permutation symbol ε_{i₁…i_N} as a tensor leaf node in 2, 3, and 4 spatial dimensions. New `levi_civita(dim)` factory in `tensor_std.h`. Full visitor coverage. 15 new tests.

### 1. Rank is fixed by dim
Unlike `identity_tensor` (which takes both dim and rank for the rank-2R minor identity at any even rank), ε is intrinsically tied to its dimension: ε in N dimensions is *always* rank-N. The constructor takes only `dim`; rank is computed.

### 2. Sign convention from parity counting, not from tmech
The data-layer evaluator (`tensor_data_levi_civita`) computes components from permutation parity directly rather than calling `tmech::levi_civita`. Reason: **tmech's dim-3 formula `(j-i)(k-j)(i-k)/2` evaluates to −1 for the identity permutation (i,j,k)=(0,1,2)**, which is the opposite of the standard sign convention (the one for which `det(I) = ε_{ijk} I_{1i} I_{2j} I_{3k}` evaluates to +1).

tmech's dim-2 (`j - i`) and dim-4 (`(j-i)(k-i)(l-i)(k-j)(l-j)(l-k)/12`) formulae are consistent with the standard convention — only dim-3 is flipped. A direct parity implementation avoids the silent trap, is no harder to read, and is independent of tmech's future changes.

### 3. Dim-4 construction allowed, evaluation rejected
`tensor_data_eval_up_unary` is instantiated with `MaxDim=3` (`numsim_cas_type_traits.h:70`). Constructing `levi_civita_tensor(4)` is allowed symbolically — useful inside expressions that collapse before numerical evaluation — but a bare numerical apply throws `evaluation_error`. Lifting MaxDim is a separate concern. Pinned by test (`EvalDim4ThrowsDueToMaxDimCeiling`).

## Critical-review iteration

Two real bugs caught during self-review:

1. **Wrong sign on dim 3.** Initial implementation delegated to `tmech::levi_civita` directly. The dim-3 evaluator test then reported all odd permutations coming back as +1 instead of −1. Tracked the formula in `tmech/levi_civita_symbol_meat.h` and confirmed tmech uses the opposite convention. Switched to a parity-counting implementation and added the antisymmetry test that hits all (i,j,k) with i≠j.

2. **Dim-4 evaluation silently corrupted memory.** Initial dim-4 evaluation test threw `dim > MaxDim || dim == 0`. The MaxDim=3 ceiling in the eval framework is a pre-existing limitation, not LC-specific — but it would silently produce nonsense if we'd allowed the LC eval to fall through. Now the `mismatch` path catches it explicitly, the contract is pinned by test, and the dim-4 limitation is documented in the node's doxygen.

## Coverage (15 tests)

- Construction at dim 2/3/4
- Plain print (`eps{N}`) and LaTeX print (`\varepsilon^{(N)}`)
- Structural equality: same-dim hash equality, cross-dim distinction, distinct-from-identity at same dim
- Eval dim 2: skew unit `[[0, 1], [-1, 0]]`
- Eval dim 3: all 6 even-permutation entries, all 6 odd-permutation entries, repeated-index zeros
- Eval dim 3 antisymmetry across all (i,j,k) with i≠j
- Eval dim 4 throws
- Differentiation of `levi_civita + A` w.r.t. `A` is valid (ε branch is dropped)
- Rebuild round-trip
- contains_expression leaf behaviour

## Out of scope

- Cross product / determinant / curl identities involving ε — those need symbolic inner-product machinery to express cleanly; the leaf node lands first, the algebra rules can follow.
- Bumping `MaxDim` to 4+ to enable dim-4 numerical evaluation — pre-existing limitation, separate concern.

Stacked on #211 (`85-numerical-diff-checks`). Per the workflow used for #137/#138/#135/#85, this is not being merged yet — under review.

Closes #34.


petlenz (commented):
## Critical review — 1 real bug, 2 doc inaccuracies, 1 weak test

### 1. [Real, fix on this PR] Construction-time accepts invalid dimensions

`levi_civita(dim)` and `make_expression<levi_civita_tensor>(dim)` accept dim ∈ {0, 1, 5, 6, …} without complaint — they construct a symbolic node that explodes only at evaluation. `tensor_data_levi_civita::mismatch()` throws `evaluation_error` when the user finally tries to evaluate, but that's a long way from the actual mistake.

Construction-time validation is consistent with the rest of the codebase (see `tensor_functions.h:309` for `permute_indices`'s rank-size validation pattern). Throw `invalid_expression_error` from the constructor with the supported-dimension range. Mirrors the `inv()` rank-2-only rejection from #192.

### 2. [Doc inaccuracy] References to `tmech::levi_civita`

The doxygen says:
> Dimensions 2, 3, and 4 are supported (matching `tmech::levi_civita`'s coverage).

But the data-layer implementation was rewritten to use parity counting precisely because `tmech::levi_civita`'s 3D sign convention is wrong. So the "matching tmech's coverage" claim is misleading — we share dim coverage but NOT semantics. Tighten the wording.

### 3. [Doc inaccuracy] `levi_civita<3>()` syntax doesn't exist

The "Construction" section says:
> From user code: `levi_civita(dim)` (free function in `tensor_std.h`) or `levi_civita<3>()` if the dimension is a compile-time constant.

There is no templated overload. `levi_civita<3>()` won't compile. Remove the false claim or actually add the templated overload. Removing is simpler — the runtime form is fine.

### 4. [Weak test] `DiffWithRespectToTensorIsZeroAccumulator` only checks `is_valid()`

The test name says "Zero Accumulator" but the assertion is just `EXPECT_TRUE(d.is_valid())` — it doesn't verify the ε contribution actually drops out. Strengthen by evaluating `diff(eps, A)` directly and asserting it's `tensor_zero` (or by checking that `diff(eps + A, A) == diff(A, A)` structurally / numerically).

### Out of scope

- Adding the templated `levi_civita<3>()` form — not necessary; runtime form is enough.
- Bumping MaxDim to enable dim-4 evaluation — separate concern.
- Substitution test for ε passing through `substitute()` — minor, the rebuild visitor passes ε through unchanged. Could add.

Fixing items 1, 2, 3, 4 now.

Issue #214, opened 2026-05-27
Add string parser for numsim_cas expressions (PEGTL, opt-in)


petlenz:
## Context

`numsim_cas` builds expressions exclusively through C++ factory functions and operator overloads. There is no path from a string (`"sin(x) + trace(A)"`) to an `expression_holder`, which blocks:

- Round-tripping serialised expressions
- Accepting expressions from config files, REPL prompts, or any non-C++ caller
- Test infrastructure that compares parsed-vs-built ASTs

This issue tracks adding a string parser that produces holders for all three domains (scalar, tensor, tensor_to_scalar).

## Proposed solution

**In-tree, opt-in component** behind `cmake -DNUMSIM_CAS_BUILD_PARSER=ON` (default OFF). Built as a separate library target `NumSim_CAS::Parser` so default consumers pay zero compile cost.

**Implementation**: PEGTL (`taocpp/PEGTL` 3.2.8, BSL-1.0, header-only, fetched via FetchContent). Picked over hand-rolled (for offloading lexer/grammar maintenance) and over lexy (for stability — PEGTL has 10+ years, semver, multi-maintainer; lexy has bus factor 1).

**Public API** in `include/numsim_cas/parser/`:

```cpp
namespace numsim::cas::parser {

class symbol_table { /* declare-or-lookup, on-the-fly */ };

using parsed_expression = std::variant<
  expression_holder<scalar_expression>,
  expression_holder<tensor_expression>,
  expression_holder<tensor_to_scalar_expression>>;

parsed_expression parse(std::string_view source, symbol_table&);
expression_holder<scalar_expression>          parse_scalar(...);
expression_holder<tensor_expression>          parse_tensor(...);
expression_holder<tensor_to_scalar_expression> parse_t2s(...);

}
```

**Grammar (v1)**:

- Numbers (int and double, locale-independent via `std::from_chars`)
- Identifiers `[a-zA-Z_][a-zA-Z0-9_]*`
- Bare ident → scalar variable (declared on first use)
- `Name{rank=R, dim=D}` → tensor declaration on first use; bare `Name` afterwards
- Binary ops: `==`, `!=`, `<`, `<=`, `>`, `>=`, `+`, `-`, `*`, `/`, `^`
- Unary `-`
[...]

PR #216, opened 2026-05-27
Parser phase 2: PEGTL grammar (#214) — landing incrementally on this branch

petlenz:
Stacked on [#215](https://github.com/NumSim-Stack/numsim-cas/pull/215). Phase 2 of the parser implementation, broken into sub-commits so each chunk is independently reviewable.

## Current state: phase 2a landed

**Phase 2a** (first commit): PEGTL FetchContent + minimal scalar arithmetic grammar. Numbers, identifiers, parens, `+ - * /` with correct precedence. 17 new tests.

## Coming on this branch

- **Phase 2b**: power (`^`, right-assoc), unary minus, six comparison operators. ~10 tests.
- **Phase 2c**: scalar function calls (sin, cos, ..., max, min, if_then_else, Macauley family). ~15 tests.
- **Phase 2d**: tensor declarations (`A{rank=2,dim=3}`), tensor/t2s functions (trace, det, norm, trans, inv, dev, sym, vol, skew, dot), contraction (inner_product, outer_product with `[i,j]` bracket-list indices). Cross-domain coercion table per the plan. ~10 tests.

Each phase will be one commit on this branch. PR can be reviewed commit-by-commit.

## Phase 2a — what's in this first commit

- **CMake**: FetchContent for taocpp/PEGTL 3.2.8 (only fetched when `NUMSIM_CAS_BUILD_PARSER=ON`). PEGTL PRIVATE-linked so it stays out of public headers.
- **Public API** `include/numsim_cas/parser/parser.h`: `parsed_expression` variant, `parse()`, plus three convenience wrappers (`parse_scalar`/`tensor`/`t2s`).
- **Private grammar** `src/numsim_cas/parser/grammar.h`: numbers (int + decimal), identifiers, primary, multiplicative and additive levels via **per-op tail rules** (`mul_tail_star`, `add_tail_plus`, …). The "single tail with sor-over-ops" approach was tried and abandoned — a pending-op variable gets overwritten when nested operator firings happen before the outer tail action. Per-op rules let each action specialisation dispatch statically. Comment in the grammar documents the trap.
- **Actions** `src/numsim_cas/parser/actions.h`: value stack of variants, operand actions push exactly one value, tail actions pop two and combine via the operator overload on holder types.
- **Translation** `src/numsim_cas/parser/parser.cpp`: `parse()` catches PEGTL's `parse_error` and translates to our `syntax_error` with snippet rendering. The three convenience wrappers `std::get_if` and throw `type_mismatch_error` on the wrong domain.

## Verification

- `cmake -DNUMSIM_CAS_BUILD_PARSER=OFF`: **1736 tests pass**, no PEGTL fetched.
- `cmake -DNUMSIM_CAS_BUILD_PARSER=ON`: **1771 tests pass** (35 parser tests = 18 from phase 1 + 17 from phase 2a).
- `clang-format-18 --Werror` clean on every new file.

## Locked-in contracts (phase 2a)

- Subtraction left-associative: `10 - 3 - 2 == 5`.
- Division left-associative: `100 / 10 / 2 == 5`.
- `*` binds tighter than `+`: `2 + 3 * 4 == 14`.
- Same identifier twice resolves to the same holder: `x + x` evaluates to `2x`, not `x + freshly-x`.
- Whitespace is skipped freely between tokens.
- Trailing garbage throws.

### DIFF
--- a/.github/workflows/clang-tidy-check.yml
+++ b/.github/workflows/clang-tidy-check.yml
@@ -25,6 +25,7 @@ jobs:
           -DCMAKE_BUILD_TYPE=Debug
           -DNUMSIM_CAS_BUILD_TESTS=OFF
           -DNUMSIM_CAS_BUILD_EXAMPLES=OFF
+          -DNUMSIM_CAS_BUILD_PARSER=ON
 
       - name: Run clang-tidy
         run: |
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,14 +176,40 @@ endif()
 # Optional parser sub-library (issue #214)
 # -----------------------------
 # When enabled, builds NumSim_CAS_Parser as a separate static/shared
-# library target. Phase 1: error types + symbol_table (no PEGTL yet).
-# Phase 2 will add parser.cpp and link taocpp::pegtl. Keeping the
-# library skeleton in place from phase 1 onwards lets downstreams
-# adopt the dependency once without churn.
+# library target. Phase 1 shipped error types + symbol_table.
+# Phase 2 adds the PEGTL grammar + actions + top-level parse() entry
+# points.
 if(NUMSIM_CAS_BUILD_PARSER)
+  # taocpp/PEGTL — header-only, pulled from upstream when the parser
+  # is opted in. Pinned to a stable tag rather than a moving branch
+  # so CI is deterministic across reruns.
+  find_package(pegtl QUIET)
+  if(NOT pegtl_FOUND)
+    FetchContent_Declare(
+      pegtl
+      GIT_REPOSITORY https://github.com/taocpp/PEGTL.git
+      GIT_TAG        3.2.8
+    )
+    set(PEGTL_BUILD_TESTS    OFF CACHE BOOL "" FORCE)
+    set(PEGTL_BUILD_EXAMPLES OFF CACHE BOOL "" FORCE)
+    set(PEGTL_INSTALL        OFF CACHE BOOL "" FORCE)
+    FetchContent_MakeAvailable(pegtl)
+    # Suppress warnings from third-party PEGTL headers.
+    get_target_property(_pegtl_real taocpp::pegtl ALIASED_TARGET)
+    if(NOT _pegtl_real)
+      set(_pegtl_real taocpp::pegtl)
+    endif()
+    get_target_property(_pegtl_inc ${_pegtl_real} INTERFACE_INCLUDE_DIRECTORIES)
+    if(_pegtl_inc)
+      set_target_properties(${_pegtl_real} PROPERTIES
+        INTERFACE_SYSTEM_INCLUDE_DIRECTORIES "${_pegtl_inc}")
+    endif()
+  endif()
+
   add_library(NumSim_CAS_Parser
     src/numsim_cas/parser/parse_error.cpp
     src/numsim_cas/parser/symbol_table.cpp
+    src/numsim_cas/parser/parser.cpp
   )
   add_library(NumSim_CAS::Parser ALIAS NumSim_CAS_Parser)
 
@@ -195,6 +221,7 @@ if(NUMSIM_CAS_BUILD_PARSER)
 
   target_link_libraries(NumSim_CAS_Parser
     PUBLIC NumSim_CAS::NumSim_CAS
+    PRIVATE taocpp::pegtl
   )
 
   target_compile_features(NumSim_CAS_Parser PUBLIC cxx_std_23)
--- a/docs/index.md
+++ b/docs/index.md
@@ -93,3 +93,4 @@ auto dd = diff(d, X);    // det(X) * inv(trans(X))
 | [Tensor-to-Scalar Domain](tensor-to-scalar.md) | 13 node types, trace/det/norm functions, simplifiers |
 | [Cross-Domain Interactions](cross-domain.md) | Cross-domain nodes, expression lifecycle, build system |
 | [Differentiation](differentiation.md) | Symbolic differentiation rules with mathematical notation |
+| [String Parser](parser.md) | PEGTL-based parser: grammar, type system, error catalogue, build flag |
--- a/docs/parser.md
+++ b/docs/parser.md
@@ -1,0 +1,284 @@
+# String Parser
+
+The string parser converts source-code expressions into the same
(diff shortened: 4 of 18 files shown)


petlenz (commented):
## Critical review pass — phases 2a–2d

Looking honestly for real bugs and coverage gaps. Findings categorized.

### Real bugs

**1. Duplicate kv in tensor declaration silently picks the last value.**
`A{rank=2, rank=4, dim=3}` parses without error and uses `rank=4`. The grammar's `tensor_kv_list = pegtl::list<tensor_kv, ...>` accepts repeated elements, and the per-kv actions overwrite the prior pending value. This is a real user-facing bug: a typo silently takes effect. Fix: detect overwrite in the kv actions and throw.

**2. `tensor_decl` accepts rank=0 or dim=0.** The `tensor` constructor takes both verbatim; no validation at construction. A user writing `A{rank=0, dim=3}` builds a degenerate tensor that fails later at evaluation, with the error pointing at a downstream site instead of the declaration. Fix: validate in the tensor_decl action.

### Coverage gaps

**3. No test for chained comparisons.** `1 < 2 < 3` parses without complaint because comparisons live at the same precedence level and left-associate. Result: `(1 < 2) < 3` = `1 < 3` = `1` (true). Python/SymPy users may expect AND-chaining semantics (`1<2 AND 2<3`). The current behaviour is the C-style value-of-comparison interpretation. **No bug** — but should be pinned by a test so a future contributor doesn't change it expecting AND.

**4. No test for `t2s + scalar` throwing.** Parallel to the existing `tensor + scalar` test. With my same-domain rule, `trace(A) + 2` should throw `type_mismatch_error`. Currently not covered.

**5. No test for error message content.** I have tests that the right exception subclass fires for type mismatches, but not that the message names the function/operator and operands involved. A user-facing diagnostic regression would slip through.

**6. No test for empty input or `(` alone.** Edge cases at the bottom of the grammar — both should throw `parse_error`.

**7. No test for `inv(scalar)` / `trans(scalar)`.** The current registry only checks `arg_kind::tensor`; passing scalar should throw `type_mismatch_error`. Test would mirror the `trace(scalar)` test that does exist.

### Fragile but not currently a bug

**8. `parser_state.pending_rank`/`pending_dim` not reset on failed tensor_decl.** If the `}` doesn't close (handled by `if_must` → throw), the prior `rank_kv`/`dim_kv` actions have already populated pending state. The throw aborts the parse so the leftover doesn't matter, but a future refactor that catches mid-parse would surface this. Defensive fix: reset pending fields at the start of `tensor_decl` action.

**9. `function_name` and `tensor_name` are duplicate rule types.** Both are `seq<identifier_first, star<identifier_rest>>` with no actions. They're distinct types only to keep the doc story clean. Minor — leave as is; consolidating to one shared rule would obscure the intent.

### Things that are fine

- Subtraction left-associative; division left-associative ✓
- `^` right-associative (`2^3^2 == 512` locked in) ✓
- Unary minus precedence between mul and power ✓
- Mixed-domain `*` works for the 6 codebase-supported pairs; `/` for the 4 ✓
- The `can_mul` / `can_div` tables match the documented surface ✓
- Lookup-first identifier semantics for bare references to declared tensors ✓
- Source/state lifetime: `what()` is baked at construction, position fields are integers — error survives source destruction ✓

Fixing items 1, 2, 8, and adding tests for 3, 4, 5, 6, 7 next.

petlenz (commented):
## Critical review — phase 2e (bracket-list + contraction)

### Real coverage gap

**1. No end-to-end evaluator round-trip for contraction.** Every other phase's critical-review pass added an integration test that actually computed against a tmech-backed evaluator. 2e only checks `parse()` returns the right variant alternative — it doesn't verify that the parsed `inner_product(A, [3,4], B, [1,2])` computes the same VALUE as a C++-built `inner_product(A, sequence{3,4}, B, sequence{1,2})`. The highest-payoff missing test, and the one most likely to catch a real bug:
- 1-based-to-0-based conversion in `to_sequence` (my code is `s[i] = iv.indices[i] - 1`; if I had the off-by-one wrong, all my variant-only tests would still pass).
- lhs/rhs argument swap in the dispatch lambda.
- An index_list_value being held when it shouldn't.

Add a test that builds a tensor with known values via tmech, parses an `inner_product` call, evaluates, compares to the expected contraction.

### Coverage gap

**2. Malformed-bracket-list internals not tested.** `[1, 2,]` (trailing comma), `[1, ,2]` (double comma), `[1 2]` (missing comma) all throw via the existing `if_must` — but I don't pin any of them. A future regression that loosened the inner list grammar (e.g. switched to `opt<list<...>>`) would silently start accepting `[1,]`. Add three short pin tests.

### Doc fix

**3. `grammar.h` header narrative says** "Phase 2d will add tensor declarations + contraction syntax" — that line is now historically wrong since 2d added tensor decls and 2e added contraction. Fix the narrative.

### Things that ARE fine but worth flagging

- `index_list_value` operand reaching `combine_mul` / `combine_div`: my `can_mul<T, index_list_value>` predicate is false, so the else branch throws `type_mismatch_error`. The message is generic ("not defined for this combination"). It can't be triggered by parsable user input — bracket-lists only appear inside arg_list — but the error is unreachable, defensive code. Leaving it.

- `inner_product` dispatch lambda uses `std::move(lhs)` while `dot_product` uses `lhs` (no move). The codebase signatures differ (inner_product takes universal ref, dot_product takes const ref). Both correct; mildly inconsistent. Leaving as-is.

- Upper-bound validation on indices: my action checks `>= 1`, not `<= rank`. The codebase's `inner_product` will catch the out-of-range at runtime. Consistent with the parser-delegates-semantic-checks pattern (cf. tensor + tensor of mismatched ranks, also caught downstream). Leaving.

### Quick design verification

I traced these to make sure nothing slipped:
- `(a + [1, 2])`: `[` doesn't start any primary, so the inner expression fails after `+`. ✓ Throws.
- `[1, 2]` standalone: same, no primary starts with `[`. ✓ Throws (pinned by `BracketListAsToplevelExpressionThrows`).
- `f([1`: `if_must` inside `index_list_literal` makes missing `]` a hard error, propagates cleanly through the surrounding sor + arg_list + function_call. ✓
- `sin([1, 2])`: arg pushed as index_list_value; function_call's type-check rejects (sin's arg_kind is scalar). ✓ (pinned by `BracketListPassedToNonContractionThrows`).

Fixing items 1, 2, 3 now.

Issue #251, opened 2026-06-01
1.0 roadmap: foundation polish (arithmetic, calculus, writing, parsing, evaluation, annotations)


petlenz:
## Goal

Ship a stable, polished, documented 1.0 — the foundation everyone needs — without blocking on the cutting-edge features (spectral decomposition, isotropic tensor functions, symbolic integration). Those become 1.1+.

The pitch to users: **"1.0 is the symbolic CAS for elementary continuum mechanics — arithmetic, calculus, parsing, serialization, evaluation, and the annotation surface that drives symmetry-aware simplification. If you need eigenvalues or Hencky strain, wait for 1.1."**

### 1. Arithmetic & operators (already in place — needs API audit)

- `+ - * /` across all three domains (scalar, tensor, t2s) ✓
- `pow` with closed-form rules ✓
- Comparisons (`gt`, `lt`, `ge`, `le`, `eq`, `ne`) ✓
- `neg`, `abs`, `sign` ✓
- Tensor algebra: `dot`, `dot_product`, `trace`, `det`, `norm`, `inv`, `trans` ✓
- Principal invariants: `first_invariant`, `second_invariant`, `third_invariant` ✓ (#247)
- Projector family: `dev`, `sym`, `vol`, `skew` ✓
- Construction-time simplifications across all of the above ✓

**Polish needed**:
- Audit the public-API surface in `numsim_cas/*.h`: remove dead code, unify naming, lock in stable namespaces.
- Document each public function with a one-line description + at least one example.

### 2. Calculus (rank-2 properly, rank-4 deferred)

- Differentiation in all three domains ✓
- `if_then_else` family with lazy evaluation ✓ (#239, #244)
- Mixed-domain differentiation (scalar w.r.t. tensor, etc.) ✓
- Chain rule via product/quotient/composition ✓
- Rank-2 inv differentiation ✓ — **but with the symmetric-aware fix** (the rank-2 subset of #250: currently the symmetric case returns the unsymmetrized result; fix this for 1.0)
- Rank-4 inv differentiation ✗ — explicit `not_implemented_error` (#250 full); defer to 1.1

**Polish needed**:
- The rank-2 symmetric-aware inv-diff fix (this is the latent bug surfaced during the Magnus/Neudecker discussion).
- A sweep through `tensor_differentiation.cpp` and `tensor_to_scalar_differentiation.cpp` for similar symmetry-aware corners.
- Lock-in tests for every node-type derivative.

[...]

Issue #246, opened 2026-05-31
Algebra-rule folds for orthogonal / positive_definite tensor annotations


petlenz:
## Context

#228 adds the annotation *surface* — `assume_orthogonal`, `assume_positive_definite`, `assume_positive_semidefinite` and their queries (landing in #245). The annotations are now queryable but don't yet drive simplification.

This issue covers the **algebra rules** the annotations should unlock — construction-time folds and propagation through the classifier infrastructure.

### Orthogonal
- `inv(orthogonal) → trans(orthogonal)` — `R^T = R^{-1}`, avoids the general inverse evaluation path. Construction-time guard in `inv()` (alongside the existing `inv(inv(A))→A`, `inv(0)→singular_error`).
- `det(orthogonal) → ±1` — sign depends on chirality. Default to `+1` for proper rotations; emit a comment that improper orthogonals (reflections) would need a `chirality` sub-flag if that's ever needed. Construction-time guard in `det()`.
- `trans(orthogonal) ⋅ orthogonal → I` and `orthogonal ⋅ trans(orthogonal) → I` — the contraction identity. Lives in the inner-product simplifier alongside the existing `sym(A) × symmetric → ...` style rules.

### Positive definite / semidefinite
- `inv(positive_definite) → positive_definite` — the inverse of a PD tensor is itself PD (propagation, not a fold). The `inv` constructor should set `AlgKind::PositiveDefinite` on its result when the operand is PD. Same for PSD if we want to propagate the weaker property — though PSD ∩ singular is the degenerate case, so propagation through `inv` should be PD-only.
- `det(positive_definite) > 0` — propagates a `positive` scalar assumption to the result of `det()`. Needs the existing scalar-assumption manager to grow a `positive` tag (or reuse an existing one if there is one).
- `sym(F^T A F) → positive_definite` when A is positive-definite and F is full-rank — congruence transformations preserve definiteness. Tricky to detect structurally; probably defer.
- Eigenvalues of PD are positive — relevant once #226 (spectral decomposition) lands, because eigendecomposition results can then propagate a scalar `positive` assumption per-eigenvalue.

## Implementation notes

1. **Where the folds live**:
   - `inv(orthogonal)→trans(orthogonal)` and `det(orthogonal)→1`: construction-time guards in `tensor_inv` / `tensor_det` constructors (or the factory in `tensor_functions.h`).
[...]

PR #254, opened 2026-06-01
Algebra folds: inv(orthogonal)→trans + det(orthogonal)→1 (#246 α-2a)

petlenz:
Step α-2a of the 1.0 roadmap (#251). First slice of #246 — turning the just-landed orthogonal annotation (#245) into actionable simplification.

## Summary

Two construction-time folds gated on the orthogonal annotation:

- \`inv(R) → trans(R)\` when R is rank-2 orthogonal. Result carries the orthogonal annotation forward.
- \`det(R) → 1\` when R is orthogonal (proper rotation; chirality not encoded today).

Un-annotated paths build \`tensor_inv\` / \`tensor_det\` unchanged.

## Implementation

| File | Change |
|---|---|
| \`include/numsim_cas/tensor/tensor_functions.h\` | Add \`is_orthogonal\` short-circuit in \`inv()\` factory, after the rank gate. Calls \`trans()\` then inserts \`orthogonal{}\` into the result's algebra-assumption manager. |
| \`src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp\` | Add \`is_orthogonal\` short-circuit in \`det()\` after the \`det(I) = 1\` case. Returns \`tensor_to_scalar_one\`. |
| \`tests/TensorAlgebraAssumeTest.h\` | Six lock-ins in a new \`TensorAlgebraFold\` test suite. |

The annotation propagation through \`trans()\` uses the post-construction-insert pattern already in use at \`tensor_operators.h:62-82\` for \`trans(A) + (-A) → Skew\`.

## Tests (6 new, all pass)

| Test | What it locks in |
|---|---|
| \`InvOrthogonalFoldsToTrans\` | \`inv(orth R)\` produces \`permute_indices_wrapper\` (the trans), NOT \`tensor_inv\`; result has \`is_orthogonal\` true |
| \`InvOrthogonalDoesNotFireAtRank4\` | Negative — rank-4 orthogonal must still build \`tensor_inv\` (no \`trans()\` at rank-4) |
| \`InvUnannotatedDoesNotFireFold\` | Negative — un-annotated rank-2 still produces \`tensor_inv\` |
| \`DetOrthogonalFoldsToOne\` | \`det(orth R)\` produces \`tensor_to_scalar_one\` |
| \`DetUnannotatedDoesNotFireFold\` | Negative — un-annotated produces \`tensor_det\` |
| \`InvOrthogonalEvaluatesCorrectly\` | End-to-end — bind R to a 30° z-rotation, evaluate \`inv(R)\`, assert componentwise equality with R^T |

Each fold has a positive case + a negative case so an "almost right" future change (e.g. firing without checking annotation) is caught immediately.

Suite: **1261/1261** (was 1255 + 6 new).

## Chirality note

\`det(orthogonal) → +1\` assumes a proper rotation. Improper orthogonals (reflections) have det = -1; the annotation surface doesn't distinguish today. Mentioned in the source comment. A future \`chirality\` sub-tag could refine this; not in 1.0 scope.

## Roadmap reference

Second commit of milestone 1.0-α (after #252's tmech pin); next is α-2b — \`inv(PD) → PD\` annotation propagation through \`tensor_inv\`'s constructor.

## Sequencing

Branched from \`main\` (post-merges, post-#252-stack-free). Touches different files from #252 (CMakeLists.txt only) — no rebase conflict.

### DIFF
--- a/tests/TensorAlgebraAssumeTest.h
+++ b/tests/TensorAlgebraAssumeTest.h
@@ -321,17 +321,20 @@ TEST(TensorAlgebraFold, InvScalarMulOrthogonalRecursesCorrectly) {
   assume_orthogonal(R);
   auto alpha = std::get<0>(make_scalar_variable("alpha"));
   auto r = inv(alpha * R);
-  // Defensive contract: the orthogonal fold must NOT have fired on
-  // α·R, so the result is not bare trans(R) — it carries the 1/α
-  // factor structurally (typically as a tensor_scalar_mul of (1/α)
-  // and trans(R)).
-  EXPECT_FALSE(is_same<identity_tensor>(r));
-  // The inv(α·A) recursion + inv(orth) fold MUST still produce
-  // something: not a bare tensor_inv node (which would mean the
-  // recursion stalled).
-  EXPECT_FALSE(is_same<tensor_inv>(r))
-      << "inv(α·R) for orthogonal R should recurse through the "
-         "scalar_mul rule and fold the inner inv(R)";
+  // POSITIVE structural check (not negative). In the bug case where
+  // is_orthogonal(α·R) wrongly returned true, the orthogonal fold
+  // would build permute_indices_wrapper(α·R, {2,1}) — that's
+  // neither identity_tensor nor tensor_inv, so absence-only
+  // assertions would silently miss the bug. Assert presence of the
+  // 1/α scalar factor: a tensor_scalar_mul wrapping trans(R).
+  ASSERT_TRUE(is_same<tensor_scalar_mul>(r))
+      << "expected (1/α)·trans(R); bare trans/permute means the orth "
+         "fold wrongly fired on α·R without preserving the scalar factor";
+  // Belt and braces: the inner tensor of the scalar_mul is trans(R),
+  // i.e. a permute_indices_wrapper.
+  auto const &sm = r.template get<tensor_scalar_mul>();
+  EXPECT_TRUE(is_same<permute_indices_wrapper>(sm.expr_rhs()))
+      << "inner of the scalar_mul should be trans(R)";
 }
 
 TEST(TensorAlgebraFold, DetScalarMulOrthogonalIsNotOne) {
--- a/tests/TensorAlgebraAssumeTest.h
+++ b/tests/TensorAlgebraAssumeTest.h
@@ -1,10 +1,13 @@
 #ifndef TENSORALGEBRAASSUMETEST_H
 #define TENSORALGEBRAASSUMETEST_H
 
+#include <cmath>
 #include <gtest/gtest.h>
+#include <memory>
 
 #include <numsim_cas/numsim_cas.h>
 #include <numsim_cas/tensor/tensor_assume.h>
+#include <numsim_cas/tensor/visitors/tensor_evaluator.h>
 
 namespace numsim::cas {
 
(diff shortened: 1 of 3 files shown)


petlenz:
`det(orth R)` produces `tensor_to_scalar_one' why not scalar_one

petlenz:
## Critical review — α-2a algebra folds

The PR cleanly threads orthogonal-annotation propagation through `trans()` and adds the `inv(orth)→trans` and `det(orth)→1` folds. CI is green, the test suite covers the dominant positive/negative cases, and the comment-rich ordering rationale around the skew-singularity guard is well-judged.

A few things worth flagging before merge.

### Soundness compromise — `det(orthogonal) → 1` silently wrong for improper rotations

The comment in `tensor_to_scalar_functions.cpp:121-128` is honest about the limitation: improper rotations (reflections) have `det = -1` and we return `+1`. The pragmatic justification (continuum-mechanics defaults to proper rotations) is reasonable, but a silently-wrong fold is worse than no fold for users who happen to be working with mirror/reflection tensors.

No tracking issue exists for the chirality sub-tag (`gh issue list` returns only #246). Recommend filing a follow-up issue ("Add `proper_rotation` / `chirality` sub-tag to refine `det(orthogonal)` fold") and linking it from the comment so the work doesn't get lost in the 1.0 transition.

### Test gap — `inv(α·R)` and `det(α·R)` for orthogonal R

The PR reorders the `inv(α·A)` scalar_mul fold to run **after** the new orthogonal fold (`tensor_functions.h:91-100`). Today this is safe because `tensor_scalar_mul` only propagates `space`, not algebra annotations like `orthogonal` (verified in `operators/scalar/tensor_scalar_mul.h:17-29`) — so `is_orthogonal(α·R)` returns false and the orthogonal fold correctly doesn't fire.

But there's no lock-in for this contract. If a future PR adds orthogonal propagation through scalar_mul (e.g. for `α = ±1`), the fold would silently wrongly collapse `α·R → trans(R)` and `det(α·R) → 1` instead of `α^d`. Add:

```cpp
TEST(TensorAlgebraFold, InvScalarMulOrthogonalRecursesCorrectly) {
  // inv(α·R) → inv(R)/α = trans(R)/α — orthogonal does NOT propagate
  // through scalar_mul, so the orthogonal fold must NOT fire on α·R.
  // Locks in the current (correct) ordering contract.
}

TEST(TensorAlgebraFold, DetScalarMulOrthogonalIsAlphaPow) {
  // det(α·R) = α^d · det(R) = α^d, NOT 1.
}
```

### Test gap — annotation on `tensor_negative` result from skew+orthogonal path

`InvSkewOrthogonal2DStillFolds` asserts `is_same<tensor_negative>(r)` but never checks `is_orthogonal(r)`. Yet `trans()`'s skew branch (`tensor_functions.h:31-36`) explicitly inserts the orthogonal annotation onto the negated result, and `(-R)` IS orthogonal when R is. Add `EXPECT_TRUE(is_orthogonal(r))` to lock in the propagation through that branch — otherwise a refactor of the skew short-circuit could drop the insert silently.

### Edge case — post-hoc annotation on a `trans` node bypassed by `trans(trans(X))→X`

If a caller annotates the `trans` node directly rather than the leaf:

```cpp
auto tR = trans(R);              // R has no annotation
assume_orthogonal(tR);            // tR carries orthogonal
auto ttR = trans(tR);             // basis_change_imp short-circuit returns R
is_orthogonal(ttR);               // false — R was never annotated
```

The standard pattern is to annotate the leaf, so this is unlikely in practice; just worth noting that the algebra-fold annotations flow construction-time only, not retroactively through short-circuits. No code change required — could be a one-line caveat in the `trans()` doxygen.

### Style nit (ignore unless cleaning up)

`propagate_orthogonal` is captured at `tensor_functions.h:382` before the rank-check throw at line 376-378. If the throw fires, the captured bool is dead. Cheap (one boolean load); harmless.

---

Net: no blockers. The chirality follow-up issue + the two missing tests would tighten this up. Recommend merging once those land or a follow-up PR opens to address them.

tests/TensorAlgebraAssumeTest.h
```diff
@@ -189,6 +193,147 @@ TEST(TensorAlgebraAssume, AlgebraAssumptionOrthogonalToProjectorSpace) {
...
+  raw[8] = 1;
+  tensor_evaluator<double> ev;
+  ev.set(R, std::static_pointer_cast<tensor_data_base<double>>(R_data));
+  auto inv_R_data = ev.apply(inv(R));
+  ASSERT_NE(inv_R_data, nullptr);
+  // For a rotation, transpose IS the inverse: R^T(0,1) = R(1,0) = s.
+  auto const &inv_R =
+      static_cast<tensor_data<double, 3, 2> const &>(*inv_R_data).data();
+  EXPECT_NEAR(inv_R(0, 1), s, 1e-12);
+  EXPECT_NEAR(inv_R(1, 0), -s, 1e-12);
```
petlenz, line 380: **The test I added here doesn't actually catch the bug it claims to.** Trace through the bug case: if `is_orthogonal(α·R)` wrongly returns true, the orthogonal fold runs `return trans(α·R)` → produces `permute_indices_wrapper(α·R, {2,1})`. That node is neither `identity_tensor` nor `tensor_inv`, so both `EXPECT_FALSE` assertions still pass.

Replace with a positive structural check:
```cpp
EXPECT_TRUE(is_same<tensor_scalar_mul>(r))
    << "expected (1/α)·trans(R); bare trans(R) means orth fold wrongly fired on α·R";
```

The companion `DetScalarMulOrthogonalIsNotOne` (line 337) IS sound — bug case there produces `tensor_to_scalar_one` which the existing assertion correctly catches.

petlenz (commented):

My three-test addition (commit `3303f77`) addresses the major findings from the first pass: scalar_mul ordering lock-in, `det(α·R)` non-collapse, and orthogonal propagation onto `-R` in the skew+orth path. Verified the fixes compile and 1266/1266 tests pass.

One issue with my own fix that the user should know about:

### `InvScalarMulOrthogonalRecursesCorrectly` is **too weak** to catch the bug it claims to lock in

In the documented bug case (a future scalar_mul change wrongly propagating `orthogonal` through α·R), the orthogonal fold would fire and execute `return trans(std::forward<Expr>(expr))`. That call resolves to `permute_indices_wrapper(α·R, {2,1})`, which is **neither `identity_tensor` nor `tensor_inv`** — so both my `EXPECT_FALSE` assertions still pass. The test gives a false sense of coverage.

Fix: assert a positive structural property — that the result carries a scalar factor. Suggested:
```cpp
EXPECT_TRUE(is_same<tensor_scalar_mul>(r))
    << "expected (1/α) · trans(R) — bare trans(R) means orth fold wrongly fired on α·R";
// Optional belt-and-braces: check the inner is trans(R).
auto const &sm = r.template get<tensor_scalar_mul>();
EXPECT_TRUE(is_same<permute_indices_wrapper>(sm.expr_rhs()));
```

The companion `DetScalarMulOrthogonalIsNotOne` test IS sound — `det(orthogonal)→1` returns `tensor_to_scalar_one`, so the bug case would be caught by the existing `EXPECT_FALSE(is_same<tensor_to_scalar_one>(d))`. Just the `inv` companion needs strengthening.

Diff against current `main` shows the old CMakeLists.txt (pre-tmech-pin) coming back as a delete. Merging as-is would revert the #252 tmech pin. Rebase on current main before clicking merge — same caveat applies to all four α-2 PRs.

### Otherwise — ship it

The fold logic is sound, the `det(orth)→1` chirality compromise is documented, the negative cases are comprehensive, and the numerical lock-in (`InvOrthogonalEvaluatesCorrectly`) is best-in-class for this PR series. Apart from the test-strength tweak above, this is ready.

Issue #228, opened 2026-05-30
Extend tensor annotations with orthogonal / positive_definite / positive_semidefinite


petlenz:
## Problem

The tensor annotation surface (`include/numsim_cas/tensor/tensor_assume.h`) currently covers:

| Annotation | Predicate |
|---|---|
| `symmetric`, `skew` | rank-2 symmetry |
| `volumetric`, `deviatoric` | sym subspaces (pressure / shear split) |
| `minor`, `major`, `minor_major` | rank-4 symmetries (stiffness tensors) |

Three annotations missing that the continuum-mechanics use cases (per [NumSim architecture spec](#)) need:

- **`orthogonal`** — rotation tensors `R^T R = I`. Polar decomposition `F = R U`, finite-strain kinematics, frame indifference checks.
- **`positive_definite`** — left/right Cauchy-Green `B = F F^T`, `C = F^T F`, mass matrix, stiffness matrix, metric tensor. **Required for `log(A)` / `sqrt(A)` (#227, isotropic tensor functions)** — those are only real-valued for positive-definite arguments.
- **`positive_semidefinite`** — Hessian at minima, damage tensors that can degenerate to singular, contact penalty matrices.

## API sketch

Match the existing pattern in `tensor_assume.h`:

```cpp
namespace numsim::cas {

void assume_orthogonal(expression_holder<tensor_expression> const &expr);
void assume_positive_definite(expression_holder<tensor_expression> const &expr);
void assume_positive_semidefinite(expression_holder<tensor_expression> const &expr);

[[nodiscard]] bool is_orthogonal(expression_holder<tensor_expression> const &);
[[nodiscard]] bool is_positive_definite(expression_holder<tensor_expression> const &);
[[nodiscard]] bool is_positive_semidefinite(expression_holder<tensor_expression> const &);

} // namespace numsim::cas
```

Plus tags in `core/assumptions.h` matching the existing `symmetric` / `skewsymmetric` / `volumetric` / `deviatoric` shape.

## Algebra rules these unlock

Construction-time folds and inferences that the existing space-classifier infrastructure would carry:

- `inv(orthogonal) → trans(orthogonal)` — much cheaper than the general inverse path
- `det(orthogonal) → ±1` (sign depends on chirality; for proper rotations, +1)
- `det(positive_definite) > 0` — propagates `positive` to a scalar via the assumption manager
- `inv(positive_definite) → positive_definite` (the inverse of a positive-definite tensor is positive-definite)
[...]

PR #255, opened 2026-06-01
Propagate PD/PSD annotations through tensor_inv (#246 α-2b)

petlenz:
Step α-2b of the 1.0 roadmap (#251). Second slice of #246.

## Summary

Where α-2a (#254) folded the orthogonal annotation in the `inv()` factory, this slice adds **PD/PSD propagation in the `tensor_inv` constructor itself** — so `inv(PD)` is PD and `inv(PSD)` is PSD for any path that builds a `tensor_inv` (rank-2 or rank-4).

### `include/numsim_cas/tensor/wrappers/tensor_inv.h`
New section in the constructor body: read the input's algebra-assumption manager and propagate `positive_definite{}` / `positive_semidefinite{}` to the result. PD joint-inserts PSD (matches #245's convention). PD/PSD imply symmetric → reuse `detail::set_symmetric_unless_more_specific` to set the projector-space tag iff no more-specific Sym subspace was already there.

Orthogonal is **not** propagated here. At rank-2 the `inv()` factory folds `inv(orthogonal)` to `trans()` before reaching this ctor (α-2a); at rank-4 "orthogonal" isn't a standard concept.

### `include/numsim_cas/tensor/tensor_assume.h` — **latent bug fix surfaced by α-2b**
Extended `detail::set_symmetric_unless_more_specific` to recognise `Minor` / `MinorMajor` perms as "more-specific sym subspaces" at rank-4. Without this fix, PD on a rank-4 MinorMajor stiffness would silently overwrite MinorMajor with rank-2 `{Sym, AnyTrace}` — the canonical continuum-mechanics use case (PD elasticity tangents) would lose rank-4 structural info.

This helper is also called from `assume_positive_definite` and `assume_positive_semidefinite`, so the fix corrects a latent pre-existing limitation from #245 / #228. No existing tests broke because none covered the rank-4 PD case before this PR — the `InvPdRank4MinorMajorPropagates` test below is what surfaced it.

## Tests (6 new, all pass) — TensorAlgebraPropagation suite

| Test | What it locks in |
|---|---|
| `InvPdIsAlsoPd` | PD propagates; PSD follows via joint insertion; symmetric implied |
| `InvPsdStaysPsd` | PSD propagates without strengthening to PD |
| `InvUnannotatedHasNoPdAnnotation` | Negative — un-annotated input → no PD/PSD on result |
| `InvPdThroughClearSpaceStillSymmetric` | Cross-mechanism: PD propagation re-derives the Sym space tag even when input had cleared it |
| `InvPdRank4MinorMajorPropagates` | Rank-4 PD stiffness; MinorMajor space survives **and** PD propagates (the test that surfaced the helper bug) |
| `InvPdComposesWithVolumetric` | Vol subspace + PD: both survive through inv |

Suite: **1261/1261** (1255 + 6 new).

## Test-suite naming

New `TensorAlgebraPropagation` suite (distinct from α-2a's `TensorAlgebraFold`) — α-2b branched from `main` before α-2a merged, and using separate test-suite names avoids a file-merge conflict. When both PRs land, the two suites coexist in `TensorAlgebraAssumeTest.h` without overlap.

## Roadmap reference

Third commit of milestone 1.0-α; next is α-2c (`trans(orthogonal) · orthogonal → I` recognition in the tensor·tensor `mul_fn`).

## Sequencing

Branched from `main` (post-merges, no dependency on #252 or #254 — touches different files).

### DIFF
--- a/include/numsim_cas/tensor/tensor_assume.h
+++ b/include/numsim_cas/tensor/tensor_assume.h
@@ -52,12 +52,23 @@ assume_minor_major(expression_holder<tensor_expression> const &expr) {
 // must be preserved.
 
 namespace detail {
-// PD/PSD => symmetric: set the space if not already a Sym subspace.
-// Incompatible spaces (Skew, Minor, Major, MinorMajor) get overwritten.
+// PD/PSD => symmetric: set the space if not already a recognised sym
+// subspace. Two families count as "sym-or-more-specific":
+//   - Rank-2 ProjKind::Sym / Vol / Dev (the existing case).
+//   - Rank-4 Minor / MinorMajor perms (added for #246 α-2b — PD on a
+//     rank-4 stiffness with minor-major symmetry is the canonical
+//     case in continuum-mechanics elasticity tangents; without this
+//     branch the helper would overwrite MinorMajor with rank-2 Sym
+//     and lose the rank-4 structural info).
+// Incompatible spaces (Skew, plain Major-only without Minor, etc.)
+// still get overwritten.
 inline void set_symmetric_unless_more_specific(tensor_expression *e) {
   if (auto const &sp = e->space()) {
     auto kind = classify_space(*sp);
     if (kind == ProjKind::Sym || kind == ProjKind::Vol || kind == ProjKind::Dev)
+      return;
+    if (std::holds_alternative<Minor>(sp->perm) ||
+        std::holds_alternative<MinorMajor>(sp->perm))
       return;
   }
   e->set_space({Symmetric{}, AnyTraceTag{}});
--- a/include/numsim_cas/tensor/tensor_assume.h
+++ b/include/numsim_cas/tensor/tensor_assume.h
@@ -61,7 +61,12 @@ namespace detail {
 //     branch the helper would overwrite MinorMajor with rank-2 Sym
 //     and lose the rank-4 structural info).
 // Incompatible spaces (Skew, plain Major-only without Minor, etc.)
-// still get overwritten.
+// still get overwritten — but ONLY at rank-2, since `Symmetric{}` is
+// a rank-2 perm. For rank-4 with no recognised space, the helper is
+// a no-op: writing a rank-2 Sym tag to a rank-4 tensor would be a
+// structural mismatch. The user can call assume_minor_major /
+// assume_major explicitly if they want a sym subspace on a rank-4
+// input; the PD/PSD algebra-manager implication still stands.
 inline void set_symmetric_unless_more_specific(tensor_expression *e) {
   if (auto const &sp = e->space()) {
     auto kind = classify_space(*sp);
@@ -71,6 +76,10 @@ inline void set_symmetric_unless_more_specific(tensor_expression *e) {
         std::holds_alternative<MinorMajor>(sp->perm))
       return;
   }
+  // Rank gate: only rank-2 gets the Symmetric{} write. Rank-4 with no
+  // qualifying space is left untouched — see the helper doc above.
+  if (e->rank() != 2)
+    return;
   e->set_space({Symmetric{}, AnyTraceTag{}});
 }
 } // namespace detail
--- a/include/numsim_cas/tensor/wrappers/tensor_inv.h
+++ b/include/numsim_cas/tensor/wrappers/tensor_inv.h
@@ -2,6 +2,7 @@
 #define TENSOR_INV_H
 
 #include <numsim_cas/core/unary_op.h>
+#include <numsim_cas/tensor/tensor_assume.h>
 #include <numsim_cas/tensor/tensor_expression.h>
 namespace numsim::cas {
 
@@ -13,6 +14,7 @@ public:
   explicit tensor_inv(
       Expr &&_expr) // NOLINT(bugprone-forwarding-reference-overload)
       : base(std::forward<Expr>(_expr), _expr.get().dim(), _expr.get().rank()) {
+    // ── Space propagation (projector-space tag) ────────────────────
     if (auto const &sp = this->expr().get().space()) {
       const auto r = this->rank();
       if (r == 2) {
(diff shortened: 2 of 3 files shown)


petlenz:
## Critical review — α-2b PD/PSD propagation through `tensor_inv`

The propagation logic is correct in its core math (PD eigenvalues stay positive under inversion; PSD propagation stays conservative). The widening of `set_symmetric_unless_more_specific` to recognize `Minor` and `MinorMajor` is the right call for rank-4 elasticity tangents. Test coverage is solid — positive cases, negative case, rank-4 path, contradictory-input lock-in, and the volumetric composition case.

A few things worth checking.

### Concern — `set_symmetric_unless_more_specific` writes rank-2 `Symmetric` to a rank-4 result with no prior space

The helper unconditionally writes `e->set_space({Symmetric{}, AnyTraceTag{}})` when no qualifying space exists. This `Symmetric` perm is rank-2 semantics. If a user reaches the algebra-propagation block on a rank-4 input that has no existing space tag — possible if they call `assume_positive_definite(rank-4 C)` without first setting `assume_minor_major` — the rank-4 `tensor_inv` result gets a structurally-mismatched rank-2 `Symmetric` tag.

This isn't introduced by the PR — the helper has the same behavior at the existing `assume_*` call sites — but the PR extends the helper's reach to a new construction point. Two options:

1. Gate the `set_symmetric_unless_more_specific` call on `this->rank() == 2`.
2. Have the helper itself rank-gate: write `Symmetric` only for rank-2, and a rank-4-appropriate Sym variant (or no-op) for rank-4.

Either way, there should be a test for `assume_positive_definite` on rank-4 C **without** prior MinorMajor to pin the current behavior (whatever it is — the test should not silently change).

### Test ordering observation — `InvContradictorySkewAndPdResolvesToSymWithPd`

The test verifies the ctor's two-block ordering (space first, then algebra) produces consistent end-state, but the input it builds — `assume_skew(W)` followed by `assume_positive_definite(W)` — already resolves at the `assume()` site to `Sym` (line 196: `EXPECT_TRUE(is_symmetric(W));`). So the ctor doesn't actually exercise the "Skew on input, PD on input" pathological path.

The docstring acknowledges this ("handles a hypothetical caller who reaches this state differently"). To actually exercise that path, the test would need to bypass `assume_positive_definite` and inject `positive_definite{}` directly into the algebra-manager AND keep `Skew` on the space — e.g.:

```cpp
W.data()->set_space({Skew{}, AnyTraceTag{}});  // overwrite back to skew
W.data()->tensor_algebra_assumptions().insert(positive_definite{});
// Now W is in the inconsistent state the ctor must handle.
```

Without that, the test only locks in the `assume_*`-site behavior, not the ctor's defensive ordering. The comment is well-meaning but the test name overpromises.

### Nit — `input_has_psd = input_has_pd || ...contains(positive_semidefinite{})`

If `assume_positive_definite` always joint-inserts PSD (per #245 convention noted in the comment), the `input_has_pd ||` disjunct is redundant. The PR keeps it for defense-in-depth against direct-manager callers — which is the same pattern PR #259 uses, so the codebases are consistent. Fine, just calling it out.

### Style nit — comment-only opinion

The "Orthogonal is intentionally NOT propagated here" paragraph (lines 76-81 of the diff) is informative for future readers but is reasoning-by-absence. Once the rank-2 fold (α-2a) is merged, `inv(orthogonal R)` never reaches this ctor at rank-2 anyway. The rank-4 paragraph is the one that stays load-bearing. Could be trimmed; not blocking.

### Test gap — `inv(inv(PD C))` annotation transitivity

`inv(PD)` is PD. So `inv(inv(PD))` should also be PD. But the inv() factory at `tensor_functions.h` has an `inv(inv(A)) → A` fold (per the inline comment). Does that fold preserve the PD annotation back to the original C? Worth a one-line lock-in:

```cpp
TEST(TensorAlgebraPropagation, InvInvPdReducesToCWithPdIntact) {
  auto C = ...; assume_positive_definite(C);
  auto invInvC = inv(inv(C));
  EXPECT_EQ(invInvC, C);                       // structural collapse
  EXPECT_TRUE(is_positive_definite(invInvC));  // C still carries PD
}
```

---

Net: the rank-4 helper-behavior concern is the main item. Solid PR otherwise. The test improvements are nice-to-have polish.

include/numsim_cas/tensor/tensor_assume.h
```diff
@@ -52,14 +52,34 @@ assume_minor_major(expression_holder<tensor_expression> const &expr) {
...
     if (kind == ProjKind::Sym || kind == ProjKind::Vol || kind == ProjKind::Dev)
       return;
+    if (std::holds_alternative<Minor>(sp->perm) ||
+        std::holds_alternative<MinorMajor>(sp->perm))
+      return;
   }
+  // Rank gate: only rank-2 gets the Symmetric{} write. Rank-4 with no
+  // qualifying space is left untouched — see the helper doc above.
+  if (e->rank() != 2)
+    return;
```
petlenz, line 82: The rank-gate makes `assume_positive_definite()` on a rank-4 input with no prior space leave the space field empty. This is the right fix — writing rank-2 `Symmetric{}` to a rank-4 tensor was a structural mismatch. The PD algebra annotation still goes in, and `is_symmetric()` returns true via the PD path.

Worth a brief note in the call sites (`assume_positive_definite` / `assume_positive_semidefinite` doc) that the cross-mechanism Sym implication is rank-2-only on the space field; for rank-4 the user should call `assume_minor_major` or `assume_major` explicitly to get a sym subspace tag. Not blocking — just helps the next reader.

petlenz (commented):

My commit `e59ad87` addresses all three findings from the first pass:

1. **Rank-gate on `set_symmetric_unless_more_specific`** — helper now no-ops on rank ≠ 2 with no qualifying space. Lock-in via `AssumePdOnRank4WithNoSpaceLeavesSpaceUnset`.
2. **Contradictory-ctor test split** — `AssumePdAfterSkewResolvesAtCallSite` documents the call-site path; `InvCtorDefendsAgainstSkewSpaceWithPdAlgebra` (using dim 2 to bypass the odd-dim singularity guard) actually exercises the ctor's defensive ordering by injecting PD directly into the algebra manager.
3. **`InvInvPdReducesToCWithPdIntact`** — annotation transitivity through the `inv(inv(A))→A` short-circuit is now pinned.

Verified the changes compile and 1265/1265 tests pass.

### Behavior-change risk audit on the rank-gate change

The new gate changes one observable behavior: `assume_positive_definite(rank-4 C with no prior space)` previously left `{Symmetric{}, AnyTraceTag{}}` on the space field; now it leaves the space field empty. The PD algebra annotation goes in regardless.

Audited consumers of `.space()` directly (without going through `is_symmetric`):
- `tensor_differentiation.h:72` — rank-2 path only, doesn't fire for rank-4.
- `tensor_evaluator.h:132, 288, 303` — operate on projector spaces and `tensor_inv` propagation, all properly handle `nullopt`.

No regression detected. Public `is_symmetric()` still returns true for rank-4 PD (via the PD-in-algebra-manager path), so users querying via the documented API see consistent behavior.

Diff vs current `main` shows the pre-tmech-pin CMakeLists.txt coming back. Rebase to incorporate #252's pin before merging — applies to all α-2 PRs.

### Otherwise — ship it

All the propagation logic is correct, the rank-4 elasticity-tangent path (Minor/MinorMajor preservation) is the right call, and the contradictory-input ordering is now properly tested. Ready.

PR #257, opened 2026-06-01
Fold trans(orthogonal) · orthogonal → I (#246 α-2c)

petlenz:
Step α-2c of the 1.0 roadmap (#251). Third slice of #246.

## Summary

Two new construction-time folds in the tensor·tensor `mul_fn` overload:

```
R · trans(R) → I_2     (when R is rank-2 orthogonal)
trans(R) · R → I_2     (symmetric ordering)
```

This is the defining relation of orthogonal tensors (R^T R = I); folding it at construction means downstream simplification sees `identity_tensor` directly, unlocking further reductions (e.g. `inv(R · trans(R)) → I`, `det(R · trans(R)) → 1`).

## Implementation

| File | Change |
|---|---|
| `include/numsim_cas/tensor/tensor_operators.h` | New folds in `tag_invoke(mul_fn, tensor, tensor)` (~line 138). Reuses `detail::is_trans_of` (the same helper used by `trans(A) ± A → Skew`). Includes added: `identity_tensor.h`, `tensor_assume.h`. |
| `tests/TensorAlgebraAssumeTest.h` | Seven lock-ins under new `TensorAlgebraMulFold` suite. |

Gated on `rank == 2` for both operands (orthogonality is rank-2 algebra; `is_trans_of` detects the {2,1} pattern which is also rank-2-specific). Rank-4 paths fall through to the generic `mul_base` simplifier unchanged.

## Tests (7 new, all pass)

| Test | What it locks in |
|---|---|
| `TransOrthogonalTimesOrthogonalIsIdentity` | Positive: `R^T · R → identity_tensor` |
| `OrthogonalTimesTransOrthogonalIsIdentity` | Symmetric ordering: `R · R^T → identity_tensor` |
| `UnannotatedTransTimesItselfDoesNotFold` | Negative: `trans(A) · A` without orthogonal must NOT fold |
| `OrthogonalTimesUnrelatedOrthogonalDoesNotFold` | Two orthogonals, neither is trans-of-other — defensive |
| `OrthogonalSelfMultiplyDoesNotFold` | `R · R` for orthogonal R is NOT I (unless R = ±I, not promised); ensures pattern requires actual transpose |
| `TransUnannotatedTimesOrthogonalDoesNotFold` | `trans(B) · R` with B ≠ R — checks the inner-equality requirement |
| `IdentityResultMatchesNumerically` | End-to-end: 3D rotation about z by π/3, eval the fold, verify the numerical result is identity (diagonal 1s) |

Each positive case is paired with at least one negative to catch any future over-eager match. Suite: **1262/1262 pass** (1255 + 7 new).

## Composition with α-2a / α-2b

This fold relies on **α-2a's `trans()` annotation propagation** — when a user writes `trans(R)` for orthogonal R, the trans()-builder marks the result as orthogonal. Without α-2a, the fold's `is_orthogonal(lhs)` check would fail for the trans-side because the permute_indices_wrapper would have empty algebra annotation.

The fold is independent of α-2b's PD propagation through `tensor_inv`. The two slices compose cleanly: PD-annotated `inv` results don't go through the fold path (they remain as `tensor_inv` nodes), and orthogonal `inv` folds at the factory level (α-2a) before reaching the mul.

## Roadmap reference

Fourth commit of milestone 1.0-α; next is α-2d (`det(PD) > 0` scalar assumption propagation through `det()`).

## Sequencing

Branched from `main`. No file conflicts with #252 (CMakeLists.txt), #254 (inv()/det() factories), or #255 (tensor_inv.h ctor).

### DIFF
--- a/include/numsim_cas/tensor/tensor_operators.h
+++ b/include/numsim_cas/tensor/tensor_operators.h
@@ -145,6 +145,21 @@ tag_invoke(mul_fn, L &&lhs, [[maybe_unused]] R &&rhs) {
   // is_trans_of detects the {2,1} permutation pattern (rank-2 transpose
   // form). Gate on rank-2 so rank-4 paths fall through to the generic
   // mul simplifier.
+  //
+  // KNOWN MISSED SIMPLIFICATIONS:
+  // - skew+orthogonal (even-dim only; e.g. 2D rotation by π/2): trans()
+  //   has a Skew short-circuit that returns -R instead of a
+  //   permute_indices_wrapper, so `trans(R) * R` becomes `(-R) * R`.
+  //   is_trans_of fails to recognise the negated form and the fold
+  //   doesn't fire, even though `(-R)·R = -R² = I` for these inputs.
+  //   Structurally correct (the math evaluates fine); just not folded.
+  // - symmetric+orthogonal (R = R⁻¹, i.e. an involution like R = I or a
+  //   reflection): `R · R = I` directly, no transpose pattern. The
+  //   current gate requires one operand to be trans(other), so this
+  //   case isn't recognised. The negative test below
+  //   (OrthogonalSelfMultiplyDoesNotFold) intentionally pins
+  //   "no fold for generic orthogonal squared" — extending to cover
+  //   sym+orth would need a separate gate, not a relaxation of this one.
   if (lhs.get().rank() == 2 && rhs.get().rank() == 2) {
     if (is_trans_of(rhs, lhs) && is_orthogonal(lhs))
       return make_expression<identity_tensor>(lhs.get().dim(), std::size_t{2});
--- a/include/numsim_cas/tensor/tensor_operators.h
+++ b/include/numsim_cas/tensor/tensor_operators.h
@@ -7,12 +7,14 @@
 #include <numsim_cas/core/promote_expr.h>
 
 #include <numsim_cas/scalar/scalar_globals.h>
+#include <numsim_cas/tensor/identity_tensor.h>
 #include <numsim_cas/tensor/operators/tensor/tensor_add.h>
 #include <numsim_cas/tensor/simplifier/tensor_simplifier_add.h>
 #include <numsim_cas/tensor/simplifier/tensor_simplifier_mul.h>
 #include <numsim_cas/tensor/simplifier/tensor_simplifier_sub.h>
 #include <numsim_cas/tensor/simplifier/tensor_with_scalar_simplifier_mul.h>
 #include <numsim_cas/tensor/simplifier/tensor_with_tensor_to_scalar_simplifier_mul.h>
+#include <numsim_cas/tensor/tensor_assume.h>
 #include <numsim_cas/tensor/tensor_zero.h>
 #include <numsim_cas/tensor_to_scalar/tensor_to_scalar_one.h>
 #include <numsim_cas/tensor_to_scalar/tensor_to_scalar_scalar_wrapper.h>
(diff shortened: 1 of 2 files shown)


petlenz:
## Critical review — α-2c `trans(orthogonal) · orthogonal → I` fold

Tight, focused PR. The fold lives in the right place (the `tag_invoke(mul_fn, tensor, tensor)` overload, mirroring the existing `trans(A) ± A → Skew` pattern in the add/sub branches). Detection uses `is_trans_of`, which is exactly the right primitive — same-symbol check, not just orthogonality on both sides. CI is green and the negative-case coverage is the strongest of the α-2 PRs: **5 negative-case tests** explicitly pin what does NOT fold.

Two observations and one minor nit.

### Missed simplification — `trans(R) · R` when R is also skew (legitimate at even dim)

Per α-2a, R can validly be annotated both skew AND orthogonal in even dimensions (the 2D-rotation-by-π/2 case). In that scenario:

- `trans(R)` runs through `trans()`'s **skew short-circuit** (`tensor_functions.h:31-36`) and returns `-R`, not a `permute_indices_wrapper`.
- The expression is then `(-R) * R`.
- `is_trans_of(rhs=R, lhs=-R)` → false (`-R` is `tensor_negative`, not `permute_indices`).
- `is_trans_of(lhs=-R, rhs=R)` → false (same reason).
- **Fold doesn't fire**, even though mathematically `(-R) * R = -R² = I` for R = [[0,-1],[1,0]].

Not a bug — the structural form `(-R) * R` is correct, just not maximally simplified. Worth either:

1. Filing a follow-up issue for "extend trans-orth-mul fold to see through `-R` for skew-orth inputs", or
2. Adding a comment in the fold body acknowledging the limitation, so a future reader trying to extend it knows where the gap is.

### Missed simplification — symmetric+orthogonal R: R · R folds to I but the pattern doesn't recognize it

If R is symmetric AND orthogonal (e.g. R = I, or any involution like a reflection), then `R · R = I` directly without needing a trans. The fold gates on `is_trans_of(...)` which requires one operand to be a transpose-wrapper. So `R * R` doesn't fold even though mathematically it equals I in this case.

`OrthogonalSelfMultiplyDoesNotFold` documents this as intentional behavior, with the rationale "R = ±I isn't promised by the annotation". That rationale is correct for a generic orthogonal R, but it misses the sym-and-orth case where `R = R⁻¹` is forced by `R^T = R` and `R^T R = I` jointly. Worth either:

1. A separate orthogonal-and-symmetric fold (`R * R → I` when both are set), or
2. Documenting in the existing negative-case docstring that "even sym+orth doesn't fold via this rule — sym+orth means `R = R⁻¹` so `R · R = I`, but that's a different fold".

Lower priority than the skew case; just calling it out.

### Nit — `TransUnannotatedTimesOrthogonalDoesNotFold` docstring confusion

The docstring (lines 134-138 of the diff) is hard to parse:

> "Even if it did match, the orthogonal check on the non-trans side would need to look at R — which IS orthogonal. But the test name fits the gating logic: the relationship must be transpose-of-self."

The reasoning is right but the sentence structure makes it sound contradictory. Suggested rewrite:

> `trans(B) * R` where B ≠ R: the `is_trans_of(lhs, rhs)` check returns false because the inner of the transpose (B) doesn't match the other operand (R). The orthogonal annotation on R alone isn't enough; the fold requires both sides be related by transposition. Lock-in against accidentally widening the gate.

### Lock-in completeness check — result `dim` correctness

`TransOrthogonalTimesOrthogonalIsIdentity` asserts `result.get().dim() == 3u` and rank == 2. Good. The `IdentityResultMatchesNumerically` test rounds out with a numerical lock-in for both that the fold fires AND that evaluation produces the right matrix. This is best-in-class — the other α-2 PRs would benefit from copying this pattern more.

---

Net: no blockers. The skew+orthogonal missed-simplification is the only thing worth tracking as a follow-up. Recommend merge after either the comment-or-issue is added.

include/numsim_cas/tensor/tensor_operators.h
```diff
@@ -138,6 +140,32 @@ tag_invoke(mul_fn, L &&lhs, [[maybe_unused]] R &&rhs) {
     const auto rank{rhs.get().rank() + lhs.get().rank() - 2};
     return make_expression<tensor_zero>(lhs.get().dim(), rank);
   }
+  // R · trans(R) = I and trans(R) · R = I for orthogonal R (#246 α-2c).
+  // Orthogonality is a rank-2 concept (R^T R = I assumes square R) and
+  // is_trans_of detects the {2,1} permutation pattern (rank-2 transpose
+  // form). Gate on rank-2 so rank-4 paths fall through to the generic
+  // mul simplifier.
+  //
+  // KNOWN MISSED SIMPLIFICATIONS:
```
petlenz, line 149: If/when someone takes a swing at extending these two missed simplifications, both could be addressed by adding two more pattern matches before the existing `is_trans_of(...)` checks:
```cpp
// skew+orthogonal: trans(R)·R via the skew short-circuit becomes (-R)·R = -R² = I.
if (is_same<tensor_negative>(lhs) && lhs.template get<tensor_negative>().expr() == rhs && is_skew(rhs) && is_orthogonal(rhs))
  return make_expression<identity_tensor>(rhs.get().dim(), 2);
// sym+orth: R·R = I when R is both sym and orthogonal (involution).
if (lhs == rhs && is_symmetric(lhs) && is_orthogonal(lhs))
  return make_expression<identity_tensor>(lhs.get().dim(), 2);
```
Mirror cases for swapped operands. Not for this PR; just sketching what the extension looks like so the next contributor has a starting point. The current PR's gate is correctly the conservative path.

petlenz (commented):

My commit `cd9866a` is comment-only — extended the fold body with `KNOWN MISSED SIMPLIFICATIONS` documenting the skew+orth (trans short-circuit returns -R) and sym+orth (involution R·R=I, no transpose pattern) cases, and rewrote the confusing `TransUnannotatedTimesOrthogonalDoesNotFold` docstring. Verified 1263/1263 tests pass; no code-behavior change.

### Nothing new — clean PR

This is the smallest and most surgical of the four α-2 folds. The negative-case test coverage (5 explicit `EXPECT_FALSE` tests) is the strongest in the series and was already a model the other PRs could learn from in the first-pass review. Nothing additional to address.

Same caveat as the other α-2 PRs — the diff vs current `main` shows the pre-tmech-pin CMakeLists.txt. Rebase before merging.

Issue #250, opened 2026-05-31
inv() differentiation: rank-4 Magnus + symmetry-aware paths at rank-2 and rank-4


petlenz:
## Context

#248 / #249 added rank-4 support to `inv()` and `tensor_inv` (construction, evaluator dispatch, printer, rebuild, substitution). Differentiation is the one visitor not wired for rank-4 — `src/numsim_cas/tensor/visitors/tensor_differentiation.cpp:245-264` throws `not_implemented_error` for `rank != 2`.

During design review of the rank-4 derivative, a broader issue surfaced: the **existing rank-2 path is also incomplete** for symmetric operands. The Magnus/Neudecker chain rule treats components as independent variables; for a symmetric or minor-symmetric tensor, the chain rule must symmetrize over the constraint group. The current rank-2 implementation uses the unsymmetrized form, which is correct only when the differentiating variable is itself "free" (no symmetry constraint propagated through the chain).

This issue now covers **all six annotation-aware paths** for `d(inv(A))/dX`, including the rank-2 cases.

## Why symmetry matters

For a symmetric rank-2 tensor $A$, components $A_{12}$ and $A_{21}$ are the **same scalar**, not two independent ones. Varying $\partial / \partial A_{kl}$ also varies $A_{lk}$. The chain rule for $\partial(A^{-1}) / \partial A$ must encode that.

Same at rank-4: for minor-symmetric $C_{ijkl}$ (e.g. elasticity tangent), $C_{ijkl} = C_{jikl} = C_{ijlk}$; varying one component varies all minor-equivalent components.

The standard Magnus/Neudecker formula treats every component independently. To get the constraint-respecting derivative, symmetrize over the symmetry group.

## Six paths to implement

| Rank | Annotation on `A` | Group | Status today |
|---|---|---|---|
| 2 | (none, general) | $\{id\}$ | ✓ existing rank-2 code |
| 2 | `Symmetric` | swap $(k \leftrightarrow l)$ | ✗ uses general formula (latent bug) |
| 2 | `Skew` (even-dim) | antisym swap $(k \leftrightarrow l)$ | ✗ uses general formula |
| 4 | (none, general) | $\{id\}$ | ✗ throws `not_implemented` |
| 4 | `Minor` | 4-element pair-swap group | ✗ throws `not_implemented` |
| 4 | `MinorMajor` | 8-element pair-swap + major-swap group | ✗ throws `not_implemented` |

[...]

PR #259, opened 2026-06-02
Propagate positivity from PD/PSD to det() result (#246 α-2d)

petlenz:
Step α-2d of the 1.0 roadmap (#251). Fourth and final slice of #246's algebra-fold milestone.

## Summary

`det()` factory now annotates the resulting `tensor_det` node based on the input's PD/PSD annotation:

| Input annotation | Result tags inserted |
|---|---|
| `positive_definite` | `positive`, `nonnegative`, `nonzero`, `real_tag` |
| `positive_semidefinite` | `nonnegative`, `real_tag` (NOT positive — could be zero singular case) |
| (none) | no propagation |

## Implementation

| File | Change |
|---|---|
| `src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp` | New propagation block in `det()`'s generic return path. Includes added: `tensor_assume.h`. |
| `tests/TensorAlgebraAssumeTest.h` | Five lock-ins in new `TensorAlgebraDetPropagation` suite. |

This is the **first cross-domain assumption propagation in the codebase**: PD on a `tensor_algebra_assumption_manager` propagates to `positive` on a `numeric_assumption_manager`. Both managers live on the `expression` base class, so direct manager access works without a new helper. Mirrors `scalar_assume.h`'s joint-insertion convention.

The propagation sits in the **generic return path**, after all structural folds (`det(I)=1`, `det(0)=0`, `det(orthogonal)=1`, `det(inv(A))=1/det(A)`, `det(α·A)=α^d·det(A)`, `det(tensor_mul)=∏det`, `det(trans(A))=det(A)`). Those folds short-circuit to constants or recursive calls — the recursive paths re-derive positivity automatically by re-calling `det()`.

## Tests (5 new, all pass)

| Test | What it locks in |
|---|---|
| `DetPdIsPositive` | PD → full positive joint (positive, nonnegative, nonzero, real_tag) |
| `DetPsdIsNonnegativeNotPositive` | PSD → nonnegative + real_tag, but NOT positive/nonzero (could be 0 in singular case) |
| `DetUnannotatedHasNoPositivity` | Negative — un-annotated input produces no positivity tags |
| `DetSkewDoesNotFirePropagation` | Skew (not PD/PSD) doesn't trigger propagation |
| `DetPdConstantFoldsBypassThePropagationPath` | `det(I) → tensor_to_scalar_one` constant still fires unchanged |

Suite: **1260/1260 pass** (1255 + 5 new).

## Why direct manager access

`scalar_assume.h::is_positive` and `assume(expr, positive{})` are typed for `expression_holder<scalar_expression>`. My result is `expression_holder<tensor_to_scalar_expression>` — different type, no implicit conversion. Both inherit `assumptions()` from the `expression` base class, so directly reading/writing `result.data()->assumptions()` works.

A cleaner long-term fix would be to add t2s-typed `assume(t2s_expr, tag)` / `is_positive(t2s_expr)` overloads. Filed as known follow-up — out of α-2d scope.

## Roadmap reference

**Closes milestone 1.0-α's algebra-fold phase**. All four α-2 sub-PRs:
- α-2a (#254): `inv(orthogonal) → trans` + `det(orthogonal) → 1`
- α-2b (#255): PD/PSD propagation through `tensor_inv` ctor
- α-2c (#257): `trans(orthogonal) · orthogonal → I`
- α-2d (this PR): `det(PD/PSD)` positivity propagation

Next milestone is 1.0-β starting with the rank-2 symmetric-aware inv differentiation (subset of #250).

## Sequencing

Branched from `main`. No file conflicts with #252 (CMakeLists), #254 (tensor_functions.h + det fold in same .cpp but different return path), #255 (tensor_inv.h), or #257 (tensor_operators.h).

Note: this PR modifies the same `det()` function as #254 (which added the `det(orthogonal) → 1` fold). They touch different code regions (#254's fold is near the top after the identity check; this PR's annotation is at the bottom generic return path) so should auto-merge cleanly, but verify on rebase.

### DIFF
--- a/tests/TensorAlgebraAssumeTest.h
+++ b/tests/TensorAlgebraAssumeTest.h
@@ -381,6 +381,141 @@ TEST(TensorAlgebraFold, InvOrthogonalEvaluatesCorrectly) {
   EXPECT_NEAR(inv_R(2, 2), 1.0, 1e-12);
 }
 
+// ─── #246 α-2d: det(PD) > 0 scalar-assumption propagation ──────────
+
+TEST(TensorAlgebraDetPropagation, DetPdIsPositive) {
+  // det of a PD tensor is strictly positive — product of positive
+  // eigenvalues. The propagation lives in det()'s generic return
+  // path; folds for identity/zero/orthogonal/etc. don't go through
+  // this annotation (those produce constants that are trivially
+  // positive or zero by construction).
+  auto C = std::get<0>(make_tensor_variable(std::tuple{"C", 3, 2}));
+  assume_positive_definite(C);
+  auto d = det(C);
+  // Direct manager access — scalar_assume.h's is_positive is typed
+  // for scalar_expression, not t2s; both inherit assumptions() from
+  // the expression base class.
+  auto const &a = d.data()->assumptions();
+  EXPECT_TRUE(a.contains(positive{}));
+  // Joint insertions per scalar_assume convention:
+  EXPECT_TRUE(a.contains(nonnegative{}));
+  EXPECT_TRUE(a.contains(nonzero{}));
+  EXPECT_TRUE(a.contains(real_tag{}));
+}
+
+TEST(TensorAlgebraDetPropagation, DetPsdIsNonnegativeNotPositive) {
+  // det of PSD is ≥ 0 — could be zero (singular case). Propagate
+  // nonnegative but NOT positive / nonzero.
+  auto H = std::get<0>(make_tensor_variable(std::tuple{"H", 3, 2}));
+  assume_positive_semidefinite(H);
+  auto d = det(H);
+  auto const &a = d.data()->assumptions();
+  EXPECT_TRUE(a.contains(nonnegative{}));
+  EXPECT_TRUE(a.contains(real_tag{}));
+  EXPECT_FALSE(a.contains(positive{}))
+      << "PSD det may be zero; must not assert strictly positive";
+  EXPECT_FALSE(a.contains(nonzero{}))
+      << "PSD det may be zero; must not assert nonzero";
+}
+
+TEST(TensorAlgebraDetPropagation, DetUnannotatedHasNoPositivity) {
+  // Negative case: det of an un-annotated tensor must NOT gain
+  // positivity tags. Guards against over-eager propagation.
+  auto A = std::get<0>(make_tensor_variable(std::tuple{"A", 3, 2}));
+  auto d = det(A);
+  auto const &a = d.data()->assumptions();
+  EXPECT_FALSE(a.contains(positive{}));
+  EXPECT_FALSE(a.contains(nonnegative{}));
+  EXPECT_FALSE(a.contains(nonzero{}));
+}
+
+TEST(TensorAlgebraDetPropagation, DetSkewDoesNotFirePropagation) {
+  // Skew tensor is not PD/PSD — no propagation. Also det is in odd
+  // dimensions actually zero (caught by the existing inv-singularity
+  // guard, but that's separate). Just sanity-check the assumption
+  // doesn't leak.
+  auto W = std::get<0>(make_tensor_variable(std::tuple{"W", 3, 2}));
+  assume_skew(W);
+  auto d = det(W);
+  auto const &a = d.data()->assumptions();
+  EXPECT_FALSE(a.contains(positive{}));
+  EXPECT_FALSE(a.contains(nonnegative{}));
+}
+
+TEST(TensorAlgebraDetPropagation, DetPdConstantFoldsBypassThePropagationPath) {
+  // Sanity: det(identity) folds to tensor_to_scalar_one (a constant),
+  // not a tensor_det node — so the annotation insertion path doesn't
+  // run. The result is already known to be 1, so positivity is
+  // implicit. Just verify the fold still fires unchanged.
+  auto I = make_expression<identity_tensor>(std::size_t{3}, std::size_t{2});
+  auto d = det(I);
+  EXPECT_TRUE(is_same<tensor_to_scalar_one>(d));
+}
+
+// ─── Recursive-path gap documentation (#246 α-2d) ─────────────────
+// The propagation block in det() fires only on the terminal
+// tensor_det node. The recursive paths compose results via t2s
+// mul/div/pow which do NOT yet propagate `positive` through the
+// operation (#260, t2s op propagation through mul/div/pow).
+//
+// Status of the recursive paths:
+//   - det(α·A) — gap open. Outer composition is pow(α,d) * det(A);
+//     even though det(A) carries positive via the recursive call,
+//     the t2s mul doesn't propagate. Test below pins the current
+//     incomplete behavior.
+//   - det(inv(C)) — gap CLOSED by PR #264, which projects PD directly
+//     onto the composed 1/det(C) result at the is_same<tensor_inv>
+//     fold site. The positive lock-in for that case lives in PR #264's
+//     composition test suite (TensorAlgebraDetPropagation.
+//     DetInvPdIsPositiveViaWrapperProjection) so we don't duplicate
+//     here. Once #260 lands generically, the projection at the fold
+//     site becomes redundant but remains correct.
+//   - det(tensor_mul) — also gap-open via #260; not specifically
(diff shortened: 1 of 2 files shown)


petlenz:
## Critical review — α-2d det(PD/PSD) positivity propagation

Core mechanism is sound: insert positivity tags into the t2s result's numeric assumption manager directly. PD ⇒ `{positive, nonnegative, nonzero, real}` and PSD ⇒ `{nonnegative, real}` — matches scalar_assume.h's joint-insertion pattern. The `set_inferred()` calls are forward-compatible.

The PR is upfront about a major limitation — it only fires on the terminal `tensor_det` node, and the recursive paths (det(α·A), det(inv), det(tensor_mul)) compose through t2s ops that don't yet propagate positivity (#260). Three "documenting the gap" tests pin the current incomplete behavior. This is honest engineering, but it creates a coordination hazard with PR #264. **See below — this is the most important finding.**

### BLOCKER — `DetInvPdLosesPositivityViaComposition` directly conflicts with PR #264

PR #264 (currently on branch `alpha2-det-inv-pd-positive`, also open) closes this exact gap by projecting PD onto the result of `det(inv(.))` at the fold site:

```cpp
// From PR #264's diff to tensor_to_scalar_functions.cpp:
if (is_same<tensor_inv>(expr)) {
  auto const &inner = expr.get<tensor_inv>().expr();
  auto result = make_expression<tensor_to_scalar_one>() / det(inner);
  if (is_positive_definite(expr)) {
    auto &a = result.data()->assumptions();
    a.insert(positive{});
    // ...
  }
  return result;
}
```

PR #259's `DetInvPdLosesPositivityViaComposition` asserts:

```cpp
EXPECT_FALSE(d.data()->assumptions().contains(positive{}))
    << "Currently expected: t2s `/` doesn't propagate positive (#260)";
```

This expectation is **false after #264 merges** — the test will FAIL.

Resolution options:

1. **Merge #259 first, then update the test in #264's rebase** — extra step in the #264 merge flow.
2. **Merge #264 first, then update the test in #259's rebase** — #259's test becomes a positive lock-in for #264's behavior; flip `EXPECT_FALSE` to `EXPECT_TRUE` and update the docstring.
3. **Update the test in #259 now** to reflect the post-#264 contract (assert `positive` IS present for `det(inv(PD))`), and keep the other two "gap" tests (`DetScalarMulPdLosesPositivityViaComposition`, which is genuinely untouched by #264) as documentation of remaining gaps.

Recommend option 3 — fix in #259 before merging, so the merge order doesn't matter. The α-2b composition gap is real and #264 closes it; reflecting that in the test suite avoids landing a test that has to be flipped within days.

### Soundness — `set_inferred()` semantics on a t2s assumptions manager

The PR uses `result.data()->assumptions()` and calls `set_inferred()`. This is the **numeric** assumption manager (inherited from the expression base class), not the algebra one. The `set_inferred()` flag is documented in the t2s/scalar simplifier comments as "these are facts, not candidates for re-derivation."

What I couldn't confirm in a quick read: does any t2s simplifier or visitor today actually consume `set_inferred()` to gate its own propagation? If nothing reads it yet (it's purely forward-compat), the comment should say so explicitly. If something does read it, the docstring on `is_inferred()` should be the source of truth. Right now the comment says "a future t2s assumption propagator should treat these as already-known facts" — implies forward-compat-only, but a sentence confirming "no current consumer" would close the loop.

### Recommended additional test — `det(C)` for PD `C` carrying conflicting `negative` annotation

Defense-in-depth: what if a user does `assume_positive_definite(C)` followed by manual injection of `negative{}` via the algebra manager (unusual but possible)? The det fold blindly inserts `positive{}` based on `is_positive_definite(expr)` without checking for conflict. The expected behavior is probably "trust the algebra annotation; user contradicted themselves" — but it'd be nice to lock that in:

```cpp
TEST(TensorAlgebraDetPropagation, DetContradictoryAnnotationsTrustsAlgebraTag) {
  auto C = ...; assume_positive_definite(C);
  // User contradicts themselves — directly inject `negative` on the tensor
  // somehow (test would need to exercise whatever path allows this).
  auto d = det(C);
  EXPECT_TRUE(d.data()->assumptions().contains(positive{}))
      << "PD on tensor takes precedence; det folds optimistically";
}
```

Skip if no such path exists; otherwise pin the behavior.

### Nit — `DetScalarMulPdLosesPositivityViaComposition` test lives in the wrong file

This test belongs in a "composition gaps" or "future-propagation" file, not the algebra-assume suite. It's an `EXPECT_FALSE` that asserts the absence of a desired behavior — different semantic category from the other algebra-propagation tests in this file. When #260 lands, these `EXPECT_FALSE` tests need to flip; they'd be easier to find if grouped.

Not blocking — just thinking about maintainability post-#260.

### Comment accuracy — `if/if` vs `if/else if` rationale

The comment says the two `if`s (rather than `if/else if`) defend against "direct-manager callers who insert positive_definite{} alone". Verified by reading: `assume_positive_definite` joint-inserts PSD, so via that path the second `if` always fires when the first does. The defense is correct — the redundant `nonnegative`/`real_tag` insertion is idempotent. Just keep the comment accurate if `assume_*` ever changes.

---

Net: **the #264-conflict is a real coordination issue** — please address before merge. The other items are polish. With option 3 applied (update the test now), this can land cleanly.

petlenz (commented):

My commit `5136c1d` removed the conflicting `DetInvPdLosesPositivityViaComposition` test, updated the comment block above the surviving gap test, and re-framed `DetScalarMulPdLosesPositivityViaComposition` as a forward-looking #260 sentinel. The blocker conflict with PR #264 is resolved.

Verified 1262/1262 tests pass.

### Cross-PR consistency check

Reviewed against PR #264's actual diff:
- #264 adds `DetInvPdIsPositiveViaWrapperProjection` (the positive lock-in)
- #264 adds `DetInvPsdDoesNotPropagatePositivity` (the PSD-skip soundness sentinel)
- This PR (#259) leaves only `DetScalarMulPdLosesPositivityViaComposition` (the remaining open gap)

The three tests partition the composition surface cleanly: scalar_mul gap open, inv gap closed by direct projection, PSD gap properly skipped. No duplication, no conflict.

### Test naming nit (very minor)

`DetScalarMulPdLosesPositivityViaComposition` could be renamed `DetScalarMulPd_OpenGap_PendingIssue260` or similar to signal more clearly that it's a sentinel for a tracked future change, not a permanent contract. Today the name implies the gap is intentional; the docstring corrects this but readers scanning test output won't see the docstring. Not blocking.

Same caveat as the other α-2 PRs. Rebase on current `main` to pick up #252's tmech pin before merging.

### Ship recommendation

This is the lightest-touch of the four α-2 PRs (single function change + one tag-insertion block). The propagation logic is correct, the PSD-skip rationale is sound, and the gap documentation now properly partitions what's closed vs open. Ready.

tests/TensorAlgebraAssumeTest.h
```diff
@@ -189,6 +189,131 @@ TEST(TensorAlgebraAssume, AlgebraAssumptionOrthogonalToProjectorSpace) {
...
+//     onto the composed 1/det(C) result at the is_same<tensor_inv>
+//     fold site. The positive lock-in for that case lives in PR #264's
+//     composition test suite (TensorAlgebraDetPropagation.
+//     DetInvPdIsPositiveViaWrapperProjection) so we don't duplicate
+//     here. Once #260 lands generically, the projection at the fold
+//     site becomes redundant but remains correct.
+//   - det(tensor_mul) — also gap-open via #260; not specifically
+//     tested here (covered by composition-test PR's broader matrix).
+
+TEST(TensorAlgebraDetPropagation, DetScalarMulPdLosesPositivityViaComposition) {
```
petlenz: When #260 lands, this test's expectation will flip from `EXPECT_FALSE` to `EXPECT_TRUE` (assuming the implementation correctly propagates `positive → positive` through t2s mul/pow with the PSD-singular caveat). At that point the test name (`...LosesPositivityViaComposition`) becomes a misnomer.

Low-effort future-proofing: rename now to something that survives the flip, e.g. `DetScalarMulPdComposition_Sentinel_Issue260`. The forward-looking framing in your docstring captures this — making it explicit in the name closes the loop. Not blocking.

Issue #262, opened 2026-06-02
Cross-domain simplification: domain-agnostic constant recognition + folds across scalar / t2s / tensor


petlenz:
## Problem

Each domain (scalar, tensor, tensor_to_scalar) defines its own constant nodes and its own simplifier paths. A "1" in one domain isn't recognised as "1" in another domain without explicit unwrapping. Simplifications that would naturally apply across domain boundaries currently require domain-specific reimplementation — or are silently missed.

### Multiple "one" nodes

| Domain | "one" node |
|---|---|
| scalar | `scalar_one` |
| t2s | `tensor_to_scalar_one` |
| tensor (rank-2) | `identity_tensor(d, 2)` |
| tensor (rank-4) | `identity_tensor(d, 4)` (minor identity) |

A simplifier pattern like `is_same<scalar_one>(expr)` returns false for `tensor_to_scalar_one` even though both represent "1". Code that wants to recognise "this expression is the constant 1" has to enumerate all variants — and miss any future addition.

### Multiple "zero" nodes

| Domain | "zero" node |
|---|---|
| scalar | `scalar_zero` |
| t2s | `tensor_to_scalar_zero` |
| tensor | `tensor_zero(d, r)` (parameterised by dim, rank) |

Same problem.

### Cross-domain wrapping

`tensor_to_scalar_scalar_wrapper(scalar_one)` represents "1" as a t2s expression. But:
- `is_same<tensor_to_scalar_one>(wrapped)` is false (it's a wrapper, not the direct sentinel).
- `is_same<scalar_one>(wrapped)` is false (the holder type is t2s, not scalar).
- A simplifier asking "is this 1?" has to unwrap manually.

This came up concretely in #246 α-2a where `det(orthogonal) → tensor_to_scalar_one`. A downstream consumer that knows about scalar's `is_one` checks but not t2s's misses the simplification opportunity. The current workaround is consumer code doing `is_same<tensor_to_scalar_one>(d) || (is_same<tensor_to_scalar_scalar_wrapper>(d) && is_same<scalar_one>(d.expr()))` — verbose, error-prone.

### Multiple arithmetic patterns

`x * 1 = x` is implemented separately in:
- `scalar/simplifier/scalar_simplifier_mul.h` (for `scalar_one` in scalar mul)
- `tensor_to_scalar/simplifier/...` (for `tensor_to_scalar_one` in t2s mul)
[...]

Issue #258, opened 2026-06-01
identity_tensor should be annotated orthogonal + symmetric in its constructor


petlenz:
## Problem

`identity_tensor` is mathematically orthogonal (`I^T · I = I`, `det(I) = 1`), but the `identity_tensor` node is constructed **without the `orthogonal{}` annotation** in its algebra-assumption manager. As a result, downstream queries return false:

```cpp
auto I = make_expression<identity_tensor>(3, 2);
is_orthogonal(I); // FALSE — should be TRUE
```

This means the `inv(orthogonal) → trans(orthogonal)` fold (#246 α-2a, PR #254) doesn't fire for `inv(I)` via the algebra path. (It happens to short-circuit via the `is_same<identity_tensor>` check first, so it works — but only because of a coincidence, not annotation-driven propagation.)

It also breaks composition: the `trans(orth) · orth → I` fold (#246 α-2c, PR #257) returns an `identity_tensor` that itself isn't annotated orthogonal. So `is_orthogonal(R · trans(R))` returns false even though the result IS the identity, which IS orthogonal.

## Where this matters

1. **Composition of folds**: future folds keyed on `is_orthogonal` won't fire on identity-derived results.
2. **User intuition**: a user calling `is_orthogonal(make_expression<identity_tensor>(3, 2))` reasonably expects true.
3. **Symmetry with `is_symmetric`**: identity IS symmetric, and `is_symmetric(I)` returns... let me check via the existing `is_symmetric` logic. It reads the space tag; if identity_tensor doesn't set a space tag either, then `is_symmetric(I)` is also false today. Same defect class.

## Fix

In `include/numsim_cas/tensor/identity_tensor.h`, in the constructor, annotate the new node:

```cpp
identity_tensor(std::size_t dim, std::size_t rank)
    : base(dim, rank) {
  // Identity tensor: trivially orthogonal (I^T I = I) and trivially
  // symmetric (I^T = I). Annotate so downstream algebra-driven folds
  // and queries see it without an explicit assume_*().
  this->set_space({Symmetric{}, AnyTraceTag{}});
  this->tensor_algebra_assumptions().insert(orthogonal{});
  // Identity is also PSD (eigenvalues all 1) but not strictly PD at
  // higher rank where 0 entries exist — skip PD/PSD until clarified.
}
```

## Tests needed

- `is_orthogonal(make_expression<identity_tensor>(3, 2))` returns true.
[...]

PR #263, opened 2026-06-02
Annotate t2s constants at construction (closes #261 + H1 pre-merge)

### LINKED ISSUE #261: T2S constants (tensor_to_scalar_one/zero) should be annotated positive/nonnegative at construction
## Problem

`tensor_to_scalar_one` and `tensor_to_scalar_zero` (the t2s constant nodes) are constructed without any assumption annotations. So:

```cpp
auto one  = make_expression<tensor_to_scalar_one>();
auto zero = make_expression<tensor_to_scalar_zero>();

one.data()->assumptions().contains(positive{});      // FALSE — should be TRUE
one.data()->assumptions().contains(nonzero{});       // FALSE — should be TRUE
zero.data()->assumptions().contains(nonnegative{});  // FALSE — should be TRUE
```

`tensor_to_scalar_one` is the constant 1 — trivially `positive`, `nonnegative`, `nonzero`, `real_tag`, `integer`, `rational`, `prime` (well, 1 isn't prime, but everything else).

`tensor_to_scalar_zero` is the constant 0 — `nonnegative`, `nonpositive`, `real_tag`, `integer`, `rational`, `nonzero` is FALSE.

Same defect class as #258 (`identity_tensor` should be annotated orthogonal + symmetric). Surfaced during the α-2d critical review: `det(I) → tensor_to_scalar_one`, and the result isn't annotated `positive` even though 1 is clearly positive.

## Fix

In the constructors of `tensor_to_scalar_one` and `tensor_to_scalar_zero` (in `include/numsim_cas/tensor_to_scalar/tensor_to_scalar_one.h` / `tensor_to_scalar_zero.h`), insert the appropriate tags into `this->assumptions()` and call `set_inferred()`.

```cpp
// tensor_to_scalar_one ctor:
tensor_to_scalar_one() {
  auto &a = this->assumptions();
  a.insert(positive{});
  a.insert(nonnegative{});
  a.insert(nonzero{});
  a.insert(real_tag{});
  a.insert(integer{});
  a.insert(rational{});
  a.set_inferred();
}
```

Same for `tensor_to_scalar_zero` with the zero-appropriate set.

Consider also `scalar_one` / `scalar_zero` — same defect probably exists for the scalar constants.

## Tests

Lock-ins per constant:
- `is_positive(tensor_to_scalar_one)` returns true (via direct manager `.contains(positive{})`).
- `is_nonnegative(tensor_to_scalar_zero)` returns true.
- `is_nonzero(tensor_to_scalar_zero)` returns false.

End-to-end (composes with #260): `det(identity_tensor)` returns `tensor_to_scalar_one`, which now should have `positive` directly. Update `DetPdConstantFoldsBypassThePropagationPath` in `TensorAlgebraDetPropagation` to assert positivity on the constant result.

## Subtle: singleton sharing

If `tensor_to_scalar_one` is a singleton (via `get_t2s_one()` or similar global accessor), mutations to its assumptions affect ALL consumers. That's actually desirable here — the annotations are mathematical invariants, not user-asserted facts that could differ per instance.

If it's NOT a singleton and each `make_expression<tensor_to_scalar_one>()` builds a fresh instance, the per-instance assumption-set initialization is the right pattern.

## Out of scope

- `scalar_constant<int>` / `scalar_constant<double>` with specific values — should derive assumptions from the value (e.g., `scalar_constant{5}` is positive, integer, nonzero). Filed separately if not already.

## Related

- #258 — `identity_tensor` annotation defaults (same defect class for the tensor domain).
- #246 α-2d (#259) — surfaced this gap via `det(I) → tensor_to_scalar_one`.
- #260 (just filed) — t2s operator propagation; together with this issue completes the propagation story.

petlenz:
**Addresses the H1 finding from the 1.0-α cross-PR review** (run via code-reviewer agent across #252/#254/#255/#257/#259).

## H1 problem statement

After α-2a (#254) and α-2d (#259) both merge to main, the same \`det()\` function produces semantically inconsistent results:

\`\`\`cpp
auto C = make_tensor_variable("C", 3, 2);
auto R = make_tensor_variable("R", 3, 2);
assume_positive_definite(C);
assume_orthogonal(R);

is_positive(det(C));  // TRUE  — α-2d sets `positive` on tensor_det
is_positive(det(R));  // FALSE — α-2a returns un-annotated tensor_to_scalar_one
\`\`\`

Both are mathematically positive constants. The asymmetry is user-visible and confusing — same factory, same input class semantics, different annotation outcomes.

## Root cause (closes #261)

T2S constants (\`tensor_to_scalar_one\`, \`tensor_to_scalar_zero\`) were constructed without any assumption annotations. Every downstream fold that returned a constant had to know to insert the math-fact tags itself, or the result would silently lack them. **Annotating once at construction is the correct layer.**

## What changed

| File | Change |
|---|---|
| \`include/numsim_cas/tensor_to_scalar/tensor_to_scalar_one.h\` | Ctor inserts \`{positive, nonnegative, nonzero, real_tag, integer, rational}\` + \`set_inferred()\` |
| \`include/numsim_cas/tensor_to_scalar/tensor_to_scalar_zero.h\` | Ctor inserts \`{nonnegative, nonpositive, real_tag, integer, rational}\` + \`set_inferred()\` (NOT positive, NOT negative, NOT nonzero) |
| \`tests/TensorAlgebraAssumeTest.h\` | Two lock-ins in new \`TensorAlgebraConstants\` suite |

**Net: +70/-2 lines across 3 files**

## Tests

- \`TensorToScalarOneCarriesPositiveTags\` — all 6 tags present.
- \`TensorToScalarZeroCarriesNonnegativeNonpositive\` — 5 tags present, negative cases (positive/negative/nonzero) absent.

**1257/1257 tests pass** (1255 baseline + 2 new). No regressions — confirms the annotations don't accidentally enable incorrect simplifications.

## What this unblocks for the α-2 milestone

Once this lands + α-2a + α-2d merge, the H1 inconsistency is gone:

- \`det(orthogonal R)\` → \`tensor_to_scalar_one\` (annotated positive in ctor)
- \`det(PD C)\` → \`tensor_det(C)\` (annotated positive by α-2d propagation)
- Both \`is_positive(...)\` queries now return TRUE consistently.

## Out of scope

- **scalar_one / scalar_zero** have the same defect class. The agent-driven review specifically flagged the t2s case (where the H1 user-visible inconsistency manifests); the scalar case has no current concrete user-facing trigger. Deferred as a sibling follow-up.
- **identity_tensor / tensor_zero** annotations (#258) — similar defect class for the tensor domain. Out of scope here; #258 tracks separately.

## Suggested merge order

Land this PR BEFORE α-2a (#254) and α-2d (#259) merge — so when they land, H1 is already addressed by construction. Alternatively land in any order; the inconsistency exists only briefly if α-2a/α-2d land first.

## Related

- #261 — t2s constants annotation gap (this PR closes it).
- #258 — identity_tensor annotation gap (sibling, separate scope).
- #262 — cross-domain simplification (broader meta-issue).
- α-1 cross-PR review surfacing this finding (no issue filed, surfaced inline).

### DIFF
--- a/include/numsim_cas/tensor_to_scalar/tensor_to_scalar_one.h
+++ b/include/numsim_cas/tensor_to_scalar/tensor_to_scalar_one.h
@@ -10,7 +10,24 @@ class tensor_to_scalar_one final
     : public tensor_to_scalar_node_base_t<tensor_to_scalar_one> {
 public:
   using base = tensor_to_scalar_node_base_t<tensor_to_scalar_one>;
-  tensor_to_scalar_one() {}
+  // The constant 1 is mathematically: positive, nonnegative, nonzero,
+  // real, integer, rational. Pre-annotate so downstream queries
+  // (e.g. is_positive(det(orthogonal R))) see the right answer without
+  // relying on a separate fold to insert these tags. Closes #261 and
+  // the H1 inconsistency surfaced by the 1.0-α cross-PR review
+  // (det(orth) returned an un-annotated 1 while det(PD) carried
+  // positive — same det() function with semantically different result
+  // annotations).
+  tensor_to_scalar_one() {
+    auto &a = this->assumptions();
+    a.insert(positive{});
+    a.insert(nonnegative{});
+    a.insert(nonzero{});
+    a.insert(real_tag{});
+    a.insert(integer{});
+    a.insert(rational{});
+    a.set_inferred();
+  }
   tensor_to_scalar_one(tensor_to_scalar_one &&data) noexcept
       : base(std::move(static_cast<base &&>(data))) {}
   tensor_to_scalar_one(tensor_to_scalar_one const &data)
--- a/include/numsim_cas/tensor_to_scalar/tensor_to_scalar_zero.h
+++ b/include/numsim_cas/tensor_to_scalar/tensor_to_scalar_zero.h
@@ -11,7 +11,21 @@ class tensor_to_scalar_zero final
 public:
   using base = tensor_to_scalar_node_base_t<tensor_to_scalar_zero>;
 
-  tensor_to_scalar_zero() = default;
+  // The constant 0 is mathematically: nonnegative, nonpositive, real,
+  // integer, rational. NOT positive, NOT negative, NOT nonzero. Same
+  // rationale as tensor_to_scalar_one — pre-annotate so downstream
+  // queries see the right answer without depending on a separate fold.
+  // Closes #261 (alongside the matching annotation in
+  // tensor_to_scalar_one).
+  tensor_to_scalar_zero() {
+    auto &a = this->assumptions();
+    a.insert(nonnegative{});
+    a.insert(nonpositive{});
+    a.insert(real_tag{});
+    a.insert(integer{});
+    a.insert(rational{});
+    a.set_inferred();
+  }
   tensor_to_scalar_zero(tensor_to_scalar_zero &&data) noexcept
       : base(static_cast<base &&>(data)) {}
   tensor_to_scalar_zero(tensor_to_scalar_zero const &data)
--- a/include/numsim_cas/tensor_to_scalar/tensor_to_scalar_zero.h
+++ b/include/numsim_cas/tensor_to_scalar/tensor_to_scalar_zero.h
@@ -12,10 +12,14 @@ public:
   using base = tensor_to_scalar_node_base_t<tensor_to_scalar_zero>;
 
   // The constant 0 is mathematically: nonnegative, nonpositive, real,
-  // integer, rational. NOT positive, NOT negative, NOT nonzero. Same
-  // rationale as tensor_to_scalar_one — pre-annotate so downstream
-  // queries see the right answer without depending on a separate fold.
-  // Closes #261 (alongside the matching annotation in
+  // integer, rational. nonzero is INTENTIONALLY absent — 0 is the
+  // additive identity, not nonzero. positive and negative are likewise
+  // absent (0 is neither). The omissions are load-bearing for
+  // soundness; the matching EXPECT_FALSE checks in
+  // TensorAlgebraConstants.TensorToScalarZeroCarriesNonnegativeNonpositive
+  // lock them in. Same rationale as tensor_to_scalar_one — pre-annotate
+  // so downstream queries see the right answer without depending on a
+  // separate fold. Closes #261 (alongside the matching annotation in
   // tensor_to_scalar_one).
   tensor_to_scalar_zero() {
     auto &a = this->assumptions();
--- a/include/numsim_cas/core/assumptions.h
+++ b/include/numsim_cas/core/assumptions.h
@@ -125,6 +125,15 @@ public:
   void clear() { set_.clear(); }
   auto const &data() const { return set_; }
 
+  // Forward-compat marker for the assumption-propagation system.
+  // set_inferred() is called by assume_* helpers (scalar_assume.h) and
+  // construction-time annotations (e.g. tensor_to_scalar_one/zero) to
+  // signal "these facts are already established; an assumption
+  // propagator should treat them as known and skip re-derivation."
+  // No current reader uses inferred(); the flag is purely future-
+  // proofing for the planned propagator. Calls to set_inferred()
+  // exist across the codebase for consistency, but produce no
+  // observable behavior today.
   bool inferred() const noexcept { return inferred_; }
   void set_inferred() noexcept { inferred_ = true; }
 
--- a/tests/TensorAlgebraAssumeTest.h
+++ b/tests/TensorAlgebraAssumeTest.h
@@ -381,6 +381,43 @@ TEST(TensorAlgebraFold, InvOrthogonalEvaluatesCorrectly) {
   EXPECT_NEAR(inv_R(2, 2), 1.0, 1e-12);
 }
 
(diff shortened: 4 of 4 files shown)


include/numsim_cas/tensor_to_scalar/tensor_to_scalar_one.h
```diff
@@ -10,7 +10,24 @@ class tensor_to_scalar_one final
...
+  // annotations).
+  tensor_to_scalar_one() {
+    auto &a = this->assumptions();
+    a.insert(positive{});
+    a.insert(nonnegative{});
+    a.insert(nonzero{});
+    a.insert(real_tag{});
+    a.insert(integer{});
+    a.insert(rational{});
+    a.set_inferred();
```
petlenz, line 29: Sanity-check: confirmed via grep that `set_inferred()` is currently a **write-only marker** — nothing in `include/` or `src/` reads `assumptions().inferred()`. The call here matches scalar_assume.h's pattern and is forward-compat, but consider one of:

1. Drop the call (since nothing consumes it; can re-add when a consumer lands).
2. Add a comment block at the `core/assumptions.h:128-129` declaration noting "no current reader; future assumption propagator will use this to skip re-derivation."

Leaning toward option 2 — the cross-codebase consistency has value even without a current consumer.

include/numsim_cas/tensor_to_scalar/tensor_to_scalar_zero.h
```diff
@@ -11,7 +11,21 @@ class tensor_to_scalar_zero final
...
-  tensor_to_scalar_zero() = default;
+  // The constant 0 is mathematically: nonnegative, nonpositive, real,
+  // integer, rational. NOT positive, NOT negative, NOT nonzero. Same
+  // rationale as tensor_to_scalar_one — pre-annotate so downstream
+  // queries see the right answer without depending on a separate fold.
+  // Closes #261 (alongside the matching annotation in
+  // tensor_to_scalar_one).
+  tensor_to_scalar_zero() {
+    auto &a = this->assumptions();
+    a.insert(nonnegative{});
```
petlenz, line 26: The omission of `positive{}`, `negative{}`, and `nonzero{}` here is the load-bearing correctness decision (0 is neither sign nor nonzero). The test `TensorToScalarZeroCarriesNonnegativeNonpositive` explicitly locks in those three absences via `EXPECT_FALSE`. 

Minor: this comment block could also call out *why* `nonzero` is conspicuously missing — a future reader scanning the list might assume the omission is a bug. One sentence: `// nonzero is INTENTIONALLY absent — 0 is the additive identity, not nonzero.`

petlenz (commented):
## Critical review — pre-annotate t2s constant nodes

Clean, minimal change addressing the H1 inconsistency surfaced by the cross-PR review: `tensor_to_scalar_one` and `tensor_to_scalar_zero` now ship with their mathematical assumptions baked in at construction. The annotation lists are correct:

- `one` → `{positive, nonnegative, nonzero, real, integer, rational}`
- `zero` → `{nonnegative, nonpositive, real, integer, rational}` (notably **not** positive/negative/nonzero, which is the subtle correctness check)

Tests cover both positive and negative cases (the `EXPECT_FALSE` checks on `zero` are the most valuable part — easy to forget that 0 doesn't carry `nonzero`).

### Observation (not blocking) — `set_inferred()` has no current consumer

Grepping confirms: `set_inferred()` is called in ~12 places across `scalar_assume.h`, the assumption propagator, and `tensor_to_scalar_functions.cpp`. The matching reader `inferred()` is **never called** anywhere in `include/` or `src/`. The flag is purely forward-compat-only today.

This is consistent with how the flag is used elsewhere (PR #259, scalar_assume), so the new calls aren't out of line. But the comments in this PR ("future propagator") and in the related PRs all dance around the fact that nothing currently consumes this. Worth either:
1. A one-line clarifying comment in `core/assumptions.h` next to the flag declaration: "Forward-compat: marker for facts the assumption propagator should treat as already-known. No current reader."
2. Or filing a tracking issue for the propagator that will eventually read it.

Without that, every future contributor will independently re-derive "why am I calling this?"

This branch is based on a `main` predating the tmech pin (#252, merged Jun 2). `git diff main..origin/alpha-pre-merge-t2s-constants-annotation -- CMakeLists.txt` shows the OLD tmech block (find_package + GIT_TAG=master) — merging this PR as-is would silently REVERT the pin. Rebase on current main first.

(Same caveat applies to all the α-2 PRs: #254, #255, #257, #259. Coordinate rebases before merging the chain.)

### Soundness — copy/move ctor semantics

The pre-annotation lives in the default ctor. The copy ctor (`tensor_to_scalar_one const&`) and move ctor (`tensor_to_scalar_one&&`) delegate to the base's copy/move, which carries the assumptions over. So copies preserve the annotations — no `n_ary_tree` rebuilds will lose them. Verified by inspection; would be worth a one-line lock-in test:
```cpp
TEST(TensorAlgebraConstants, OneAnnotationsSurviveCopy) {
  auto a = make_expression<tensor_to_scalar_one>();
  auto b = a;  // shared_ptr copy — same node, same annotations trivially
  auto c = make_expression<tensor_to_scalar_one>(*a.template get<tensor_to_scalar_one>().data());  // deep copy via copy ctor
  EXPECT_TRUE(c.data()->assumptions().contains(positive{}));
}
```
Nice-to-have, not required.

---

Net: ship it after the rebase. Annotations are mathematically correct and the tests pin them well.

Issue #260, opened 2026-06-02
T2S operators (mul/div/pow/neg) should propagate positive/nonnegative through compositions


petlenz:
## Problem

Tensor-to-scalar binary operators (`*`, `/`, `pow`) and the unary `negative` don't propagate `positive` / `nonnegative` / `nonzero` / `real_tag` numeric assumptions from operands to the result. This breaks compositional reasoning about positivity:

```cpp
// All four sub-expressions are positive, but the OUTER expressions are
// not annotated positive because the operators don't propagate.

auto t = trace(PD_C);          // t is *not* annotated positive (#262 covers trace)
auto d = det(PD_C);            // d IS annotated positive (#246 α-2d landed)
auto r1 = d * d;               // r1 NOT annotated positive  (mul gap)
auto r2 = d / d;               // r2 NOT annotated positive  (div gap)
auto r3 = pow(d, 2);           // r3 NOT annotated positive  (pow gap)
auto r4 = -d;                  // r4 SHOULD be annotated negative (sign gap)
```

This was surfaced during the critical review of #259 (`det(PD)` → `positive` propagation, PR for #246 α-2d). The det() factory now correctly annotates the terminal tensor_det node, but recursive paths (`det(α·PD)` → `α^d · det(PD)`, `det(inv(PD))` → `1/det(PD)`) lose the annotation when they compose through these operators.

## Where

`src/numsim_cas/tensor_to_scalar/tensor_to_scalar_operators.h` (or similar) — the `tag_invoke(mul_fn, t2s, t2s)`, `tag_invoke(div_fn, t2s, t2s)`, and `tag_invoke(neg_fn, t2s)` overloads, plus `pow()` in `tensor_to_scalar_std.h`.

`src/numsim_cas/scalar/scalar_operators.h` — same gap likely exists for scalar operators; many scalar expressions could carry positivity that's currently dropped at each composition step.

## Propagation rules to implement

| Operation | Rule | Implies |
|---|---|---|
| `pos × pos` | result is `positive` | also nonneg, nonzero, real |
| `pos × nonneg` | result is `nonnegative` | also real |
| `nonneg × nonneg` | result is `nonnegative` | also real |
| `pos / pos` | result is `positive` | also nonneg, nonzero, real |
| `pos / nonzero-pos` | same as above (division by zero would NaN at eval) |
| `pow(pos, anything)` | result is `positive` | also nonneg, nonzero, real |
[...]

PR #264, opened 2026-06-02
Close α-2b composition gap: project PD onto det(inv(.)) result

petlenz:
## Summary

Closes the F2 finding from the cross-PR composition review on the 1.0-α milestone. α-2b annotates `tensor_inv(PD)` wrappers with `positive_definite{}`, but `det()` matches `is_same<tensor_inv>` and returns `1/det(inner)` BEFORE any PD check — leaving the wrapper's tag invisible and the composed result un-annotated despite being mathematically strictly positive.

Two-line fix at the fold site projects PD from the outer wrapper onto the composed `one / det(inner)` result. Avoids waiting on #260 (t2s div positivity propagation through pow) for this particular chain.

## Why this PR / not amend α-2d

α-2d's `Limitation` comment explicitly documented this gap. The fix is a natural extension and conceptually belongs there, but α-2d's PR (#259) is already in review with the gap-lock-in test. This PR replaces that test with the positive lock-in plus a new PSD soundness lock-in, and is stacked on the integration branch so reviewers can see the chain against a tree that has α-2b's wrapper available.

## Dependencies

- **α-2b (#255)** — provides the `tensor_inv(PD)` wrapper annotation that the new check queries.
- **α-2d (#259)** — provides the `det()` annotation-propagation pattern this PR extends.

Merge order: α-2b → α-2d → this. Retarget to `main` per the repo convention before merging.

## Soundness note: PSD intentionally NOT propagated

PSD allows `det = 0` (singular case); `1/0` is undefined. Propagating any positivity or nonneg claim on a potentially-NaN/inf value would be unsound. Locked in by `DetInvPsdDoesNotPropagatePositivity`.

## Test plan

- [x] `DetInvPdIsPositiveViaWrapperProjection` (replaces prior gap lock-in) — passes
- [x] `DetInvPsdDoesNotPropagatePositivity` (new PSD soundness lock-in) — passes
- [x] Other α-2d propagation tests unchanged — all pass
- [x] All 5 composition tests on the integration branch — pass
- [x] Full suite: 1295 / 1295 green

## What remains for #260

`det(α·PD)` and `det(tensor_mul of PDs)` chains remain open — those still need t2s mul/div/pow positivity propagation generically. `DetScalarMulPdLosesPositivityViaComposition` is kept as the open-gap sentinel.

### DIFF
--- a/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
+++ b/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
@@ -111,9 +111,27 @@ det(expression_holder<tensor_expression> const &expr) {
 
   // det(inv(A)) = 1/det(A). Routes through the t2s div operator which
   // composes via pow(rhs, -1) — produces canonical pow(det(A), -1).
+  //
+  // Closes the α-2b composition gap surfaced by the cross-PR review:
+  // α-2b annotates tensor_inv(PD) with positive_definite{}, but the
+  // recursive det(inner) call doesn't see the outer wrapper's tag,
+  // and t2s div positivity propagation (#260) is open. So PD on the
+  // outer expression has to be projected onto the composed result
+  // here directly. PD ⇒ det > 0 strict, so 1/det is also strictly
+  // positive. PSD is NOT propagated: det may be zero and 1/det is
+  // then undefined; any positivity claim would be incorrect.
   if (is_same<tensor_inv>(expr)) {
     auto const &inner = expr.get<tensor_inv>().expr();
-    return make_expression<tensor_to_scalar_one>() / det(inner);
+    auto result = make_expression<tensor_to_scalar_one>() / det(inner);
+    if (is_positive_definite(expr)) {
+      auto &a = result.data()->assumptions();
+      a.insert(positive{});
+      a.insert(nonnegative{});
+      a.insert(nonzero{});
+      a.insert(real_tag{});
+      a.set_inferred();
+    }
+    return result;
   }
 
   // det(trans(A)) = det(A). trans() builds permute_indices_wrapper with
--- a/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
+++ b/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
@@ -132,6 +132,9 @@ det(expression_holder<tensor_expression> const &expr) {
   if (is_same<tensor_inv>(expr)) {
     auto const &inner = expr.get<tensor_inv>().expr();
     auto result = make_expression<tensor_to_scalar_one>() / det(inner);
+    // REMOVE WHEN #260 LANDS: once t2s div/pow propagate positivity
+    // generically, `1 / positive` will produce a positive-annotated
+    // result on its own and this explicit projection is redundant.
     if (is_positive_definite(expr)) {
       auto &a = result.data()->assumptions();
       a.insert(positive{});
--- a/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
+++ b/src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
@@ -120,6 +120,15 @@ det(expression_holder<tensor_expression> const &expr) {
   // here directly. PD ⇒ det > 0 strict, so 1/det is also strictly
   // positive. PSD is NOT propagated: det may be zero and 1/det is
   // then undefined; any positivity claim would be incorrect.
+  //
+  // CONTRACT NOTE: the is_positive_definite check below interrogates
+  // `expr` (the OUTER tensor_inv wrapper), NOT `inner`. This couples
+  // to α-2b's tensor_inv ctor, which propagates PD from inner→wrapper.
+  // A future refactor that moves the check to `inner` would change
+  // the contract: a direct make_expression<tensor_inv>(C) that bypasses
+  // the inv() factory's ctor-side propagation would suddenly behave
+  // differently from inv(C). If you ever need to read PD from the
+  // inner, do so EXPLICITLY rather than swapping the argument here.
   if (is_same<tensor_inv>(expr)) {
     auto const &inner = expr.get<tensor_inv>().expr();
     auto result = make_expression<tensor_to_scalar_one>() / det(inner);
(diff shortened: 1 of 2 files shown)


src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
```diff
@@ -111,9 +111,27 @@ det(expression_holder<tensor_expression> const &expr) {
...
+  //
+  // Closes the α-2b composition gap surfaced by the cross-PR review:
+  // α-2b annotates tensor_inv(PD) with positive_definite{}, but the
+  // recursive det(inner) call doesn't see the outer wrapper's tag,
+  // and t2s div positivity propagation (#260) is open. So PD on the
+  // outer expression has to be projected onto the composed result
+  // here directly. PD ⇒ det > 0 strict, so 1/det is also strictly
+  // positive. PSD is NOT propagated: det may be zero and 1/det is
+  // then undefined; any positivity claim would be incorrect.
   if (is_same<tensor_inv>(expr)) {
```
petlenz, line 132: Worth a one-line architecture comment here noting the dependency on α-2b's wrapper-side PD annotation: this projection fires on `expr` (the outer `tensor_inv` node), not on `inner`. If a future refactor moved the check to `inner` it would change the contract — currently `tensor_inv(C)` carries PD via α-2b's ctor propagation, but a direct `make_expression<tensor_inv>(C)` bypassing the factory would not. Locking the contract location explicitly in the comment prevents accidental drift.

src/numsim_cas/tensor_to_scalar/tensor_to_scalar_functions.cpp
```diff
@@ -111,9 +111,27 @@ det(expression_holder<tensor_expression> const &expr) {
...
     auto const &inner = expr.get<tensor_inv>().expr();
-    return make_expression<tensor_to_scalar_one>() / det(inner);
+    auto result = make_expression<tensor_to_scalar_one>() / det(inner);
+    if (is_positive_definite(expr)) {
+      auto &a = result.data()->assumptions();
+      a.insert(positive{});
+      a.insert(nonnegative{});
+      a.insert(nonzero{});
+      a.insert(real_tag{});
+      a.set_inferred();
```
petlenz, line 144: Same `set_inferred()` observation as on PR #263 — purely forward-compat write today. Consistent with the codebase pattern; no change needed.

petlenz (commented):
## Critical review — close α-2b composition gap via PD projection on det(inv(.))

Surgical fix for the F2 architectural gap surfaced during composition testing: α-2b annotates `tensor_inv(PD C)` with `positive_definite{}`, but `det()`'s `is_same<tensor_inv>` short-circuit returns `one / det(inner)` before the bottom-of-function PD-projection runs, so the wrapper's tag is invisible to the composed result. This PR's solution — project PD from the outer wrapper directly at the fold site — is the right local fix while #260 (generic t2s div positivity propagation) remains open.

The PSD-skip soundness reasoning is correct and the sentinel test `DetInvPsdDoesNotPropagatePositivity` will catch a future #260 implementation that naively propagates nonneg through `pow(x, -1)`.

### BLOCKER — PR base is not `main`; coordinated merge needed

This PR currently targets `alpha2-composition-tests`, which itself is an integration branch (~677 lines of cross-PR composition tests, not a PR). Per the codebase's merge convention (memory rule: never merge to a stacked base; always retarget to `main` first), one of the following must happen before this can ship:

1. **Open a PR for `alpha2-composition-tests`** carrying the composition test suite, merge that to `main` after the α-2 PRs land, then retarget this PR's base to the new `main`. Diff stays small (just this fix + its lock-in tests).
2. **Fold the composition test suite into this PR**, retarget to `main`. Diff balloons to ~900 lines but everything lands together.
3. **Cherry-pick just the det() projection into a fresh branch off `main`**, accompanied by the new `DetInvPdIsPositiveViaWrapperProjection` / `DetInvPsdDoesNotPropagatePositivity` tests. The `alpha2-composition-tests` branch survives independently for whatever cross-PR scaffolding it carries.

Recommend option 1 for diff clarity. If `alpha2-composition-tests` isn't going to land as its own PR, prefer option 3.

### Soundness — PD projection on the *outer* expression

The `is_positive_definite(expr)` check looks at the algebra annotation on `expr` (the `tensor_inv` wrapper), not on `inner`. This is correct given α-2b's contract that `tensor_inv(PD)` itself carries PD. If someone bypasses α-2b and constructs a `tensor_inv` directly without the wrapper-side PD tag (e.g. via a future codepath), the projection won't fire and `det(inv(C))` will silently lose positivity. That's the intended behavior — the PR explicitly couples this fix to α-2b's wrapper-side annotation — but worth flagging in the comment so a refactor doesn't move the check to `inner` without realising it changes the contract.

### Coordinated test cleanup with PR #259

PR #259's recently-pushed fix (commit 5136c1d) removed `DetInvPdLosesPositivityViaComposition` and replaced it with a comment block referencing this PR's closure. Verified the two are consistent: #259 acknowledges the gap is closed *here*; this PR adds the positive lock-in (`DetInvPdIsPositiveViaWrapperProjection`). The merge order #259 → (rebase #264) is now safe; the previous order conflict is resolved.

### Nit — `set_inferred()` write-only marker

Consistent with the rest of the codebase but `inferred()` is never read anywhere — see the comment on PR #263 for the broader observation. No action needed here.

petlenz:
## Follow-up — base retargeting recommendation

After completing the round of code/test fixes and force-pushing rebases to all 5 sibling α-2 PRs (#254, #255, #257, #259, #263), the only remaining structural concern on this PR is its base: **`alpha2-composition-tests`**, not `main`.

Three options from the earlier review, with a recommendation:

**Option A — promote `alpha2-composition-tests` to its own PR (recommended)**
1. Merge the five α-2 PRs to `main` one at a time (per-PR auth, `--merge`).
2. Rebase `alpha2-composition-tests` on the resulting `main` — its diff will collapse to just the ~677-line composition test file (since the code changes are now in main).
3. Open a PR for that branch → `main`. Review and merge.
4. Retarget this PR (#264) base to `main` via:
   ```
   gh api -X PATCH /repos/NumSim-Stack/numsim-cas/pulls/264 -f base=main
   ```
   The diff collapses to just the `det()` PD-projection change + the two new tests.

**Option B — fold the composition tests into this PR**
Cherry-pick the composition test additions onto this branch, retarget base to `main`. Single PR but ~900 lines.

**Option C — discard `alpha2-composition-tests`, cherry-pick onto fresh branch**
Open a new branch off `main` with just this PR's content + the two new lock-in tests. Lose the composition test scaffolding entirely.

Recommend **A** — cleanest history, preserves the composition tests as a reviewable artifact, keeps this PR's diff small. The work to set it up is small (one rebase, one PR open) once the α-2 chain lands.

Ready to assist with the merge sequencing when you're ready to drive it.

Issue #282, opened 2026-06-08
Position-threading for inside-dispatch parser errors (permute_indices, inner_product, dot_product, constant factories)


petlenz:
## Context

When the parser's function-call dispatch raises an error from inside the dispatch lambda — e.g. `permute_indices()` throws `invalid_expression_error`, `to_positive_size_t` throws `type_mismatch_error` — the error has no source position, because the dispatch lambda can't see the byte offset that the `function_call` action knows.

## Concrete examples

- `parse_scalar("zero_tensor(0, 2)", syms)` raises `type_mismatch_error("zero_tensor: dim must be positive (got 0)", /*byte_offset=*/0, /*source=*/"")` — `position()` is 0, no caret, no line/column.
- `parse_tensor("permute_indices(A{rank=2, dim=3}, [1, 1])", syms)` raises `invalid_expression_error` from `tensor_functions.h` — bubbles up untranslated, also no parser-level position.

## Affected dispatch entries

- `zero_tensor`, `identity_tensor`, `levi_civita` (via `to_positive_size_t`)
- `permute_indices`, `inner_product`, `dot_product` (via library factory throws)

## Proposal

Thread the call's byte offset into the dispatch. Options:

1. **Pass `byte_offset` as a third arg to `dispatch`** — change `std::function<parsed_expression(arg_vec)>` to `std::function<parsed_expression(arg_vec, std::size_t)>` and update every entry helper. Then the helpers can re-throw with the correct position.

2. **Thread-local position slot** — the `function_call` action sets a thread-local before calling dispatch; dispatch reads it when constructing errors. Less invasive to the entry signatures but introduces global state.

3. **Catch + rethrow in the `function_call` action** — wrap the `dispatch(args)` call in a try/catch that rethrows with the call's position. Cleanest, no signature change, single integration point.

## Recommendation

Option 3. The action already has the position; one catch site retrofits all dispatch entries at once.

## Acceptance

- Parser errors raised inside any registered dispatch carry the call's position
- Existing tests still pass
- Add tests for at least one error per affected entry pinning the position info

## Origin

PR #281 pass-1 MED M2; PR #281 pass-2 LOW-3. Deferred from #281 as a wider refactor.

petlenz:
New sources for this, from #360 / PR #438 (confirmed with a parser build):

- `parse("A + W")` (dim 3 vs dim 2) throws `invalid_expression_error: tensor operator+: operands must share dim and rank (...)`
[...]

PR #330, opened 2026-07-21
Parser: add missing functions + arg-kind overload resolution

petlenz:
Fills gaps between the parser's function registry and the CAS function surface,
and adds the overload-resolution infrastructure the isotropic functions need.

## New functions
- **Principal invariants** — `first_invariant`, `second_invariant`,
  `third_invariant` (tensor → t2s).
- **Minor-basis outer products** — `otimesu`, `otimesl` (tensor, tensor → tensor).
- **Spectral accessors** — `eigenvalue(A, i)` → λ_i (t2s), `eigenvector(A, i)` →
  n_i, `eigenprojection(A, i)` → E_i (tensor), with a 0-based integer-literal
  index.

## Overload resolution (the infrastructure win)
The registry is now a `multimap`, and the `function_call` action resolves among a
name's entries by matching argument kinds (arity first, then kinds). This lets
`log`/`exp`/`sqrt` serve **both** the scalar form `log(x)` and the isotropic
tensor form `log(A)` under one name — the arg-kind resolver the registry header
had long flagged as needed. Single-entry names are unaffected, and the resolver
also unblocks the deferred mixed-domain `if_then_else` overloads (only the entries
remain to add).

Errors stay precise: wrong arity → `arity_error`; no matching kinds →
`type_mismatch_error` ("no overload of 'f' matches the given argument types").

## Tests
New `ParserFunctions` suite: each new function's result domain, the
scalar-vs-tensor overload of `log`/`exp`/`sqrt`, the spectral accessors, and both
error paths. **1608 tests pass** (no regression).

Scalar coverage was already complete, so this closes the remaining tensor/t2s
gaps. Notably it makes spectral / Hencky-style models writable in source form
(`log(C)`, `eigenvalue(A, i)`), complementing the codegen spectral work.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/numsim_cas/parser/actions.h
+++ b/src/numsim_cas/parser/actions.h
@@ -474,16 +474,14 @@ template <> struct action<grammar::function_call> {
     state.arg_marks.pop_back();
     auto arg_count = state.values.size() - mark;
 
-    // Resolve in the polymorphic registry.
+    // Resolve in the polymorphic registry. A name may carry several overloads
+    // (multimap) distinguished by argument kinds — e.g. scalar `log(x)` vs
+    // isotropic tensor `log(A)`. Pick the overload whose arg kinds match the
+    // actual arguments.
     auto const &reg = registry::function_registry();
-    auto it = reg.find(name);
-    if (it == reg.end()) {
+    auto const [cand_begin, cand_end] = reg.equal_range(name);
+    if (cand_begin == cand_end) {
       throw unknown_function_error(std::move(name), pos, state.source);
-    }
-    auto const &entry = it->second;
-    if (arg_count != entry.arg_kinds.size()) {
-      throw arity_error(std::move(name), entry.arg_kinds.size(), arg_count, pos,
-                        state.source);
     }
 
     // Collect args in source order. The stack has them in push order
@@ -496,37 +494,58 @@ template <> struct action<grammar::function_call> {
     }
     std::reverse(args.begin(), args.end());
 
-    // Type-check each arg against the entry's declared kinds before
-    // calling dispatch — dispatch uses unchecked `std::get` so a
-    // mismatch here would throw `std::bad_variant_access` rather
-    // than our nicer type_mismatch_error.
-    for (std::size_t i = 0; i < arg_count; ++i) {
-      bool ok = false;
-      switch (entry.arg_kinds[i]) {
+    auto kind_matches = [](registry::parser_value const &v,
+                           registry::arg_kind k) -> bool {
+      switch (k) {
       case registry::arg_kind::scalar:
-        ok = std::holds_alternative<expression_holder<scalar_expression>>(
-            args[i]);
-        break;
+        return std::holds_alternative<expression_holder<scalar_expression>>(v);
       case registry::arg_kind::tensor:
-        ok = std::holds_alternative<expression_holder<tensor_expression>>(
-            args[i]);
-        break;
+        return std::holds_alternative<expression_holder<tensor_expression>>(v);
       case registry::arg_kind::index_list:
-        ok = std::holds_alternative<registry::index_list_value>(args[i]);
+        return std::holds_alternative<registry::index_list_value>(v);
+      }
+      return false;
+    };
+
+    // Match on arity first, then on arg kinds — the kind match doubles as the
+    // type check (dispatch uses unchecked std::get, so the chosen overload's
+    // kinds are guaranteed to fit).
+    registry::function_entry const *chosen = nullptr;
+    bool arity_seen = false;
+    std::size_t const some_arity = cand_begin->second.arg_kinds.size();
+    for (auto it = cand_begin; it != cand_end; ++it) {
+      auto const &cand = it->second;
+      if (cand.arg_kinds.size() != arg_count) {
+        continue;
+      }
+      arity_seen = true;
+      bool all = true;
+      for (std::size_t i = 0; i < arg_count; ++i) {
+        if (!kind_matches(args[i], cand.arg_kinds[i])) {
+          all = false;
+          break;
+        }
+      }
+      if (all) {
+        chosen = &cand;
         break;
       }
-      if (!ok) {
-        throw type_mismatch_error(
-            "function '" + name + "': argument " + std::to_string(i + 1) +
-                " has wrong type for the expected signature",
-            pos, state.source);
+    }
+
+    if (chosen == nullptr) {
+      if (!arity_seen) {
+        throw arity_error(std::move(name), some_arity, arg_count, pos,
+                          state.source);
       }
+      throw type_mismatch_error("no overload of '" + name +
+                                    "' matches the given argument types",
+                                pos, state.source);
     }
 
     // Dispatch returns `parsed_expression` (3-variant); convert up to
     // the wider `parser_value` (4-variant) before pushing onto the
     // parser-internal stack.
-    auto result = entry.dispatch(std::move(args));
+    auto result = chosen->dispatch(std::move(args));
     state.values.emplace_back(std::visit(
         [](auto &&v) -> registry::parser_value { return std::move(v); },
         std::move(result)));
--- a/src/numsim_cas/parser/actions.h
+++ b/src/numsim_cas/parser/actions.h
@@ -501,6 +501,9 @@ template <> struct action<grammar::function_call> {
         return std::holds_alternative<expression_holder<scalar_expression>>(v);
       case registry::arg_kind::tensor:
         return std::holds_alternative<expression_holder<tensor_expression>>(v);
+      case registry::arg_kind::tensor_to_scalar:
+        return std::holds_alternative<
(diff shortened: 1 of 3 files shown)


tests/ParserTest.h
```diff
@@ -2005,6 +2005,80 @@ TEST(ParserErrorCoverage, UnknownSymbolErrorIsUnreachableFromParser) {
   EXPECT_TRUE(syms.has("never_seen_before"));
 }
 
+// ─── Newly registered functions (invariants, otimesu/l, isotropic, spectral)
+
+TEST(ParserFunctions, PrincipalInvariantsReturnT2s) {
```
petlenz: **[Medium — test strength] These assert the result DOMAIN, not node identity.** A dispatch wired to the wrong function would pass: I verified `first_invariant(A)` and `trace(A)` are hash-EQUAL (correct — I1 = tr — but it means the domain check can't tell them apart), so if e.g. `second_invariant` were mis-wired to `trace`, or `otimesu`/`otimesl` were swapped (both tensor-valued), the test would still be green. Add hash-equivalence assertions that pin each function to its intended node and DISTINGUISH the siblings (otimesu ≠ otimesl ≠ otimes; each invariant distinct), following the existing otimes/outer_product equivalence-lock pattern.

src/numsim_cas/parser/actions.h
```diff
@@ -496,37 +494,58 @@ template <> struct action<grammar::function_call> {
...
     std::reverse(args.begin(), args.end());
 
-    // Type-check each arg against the entry's declared kinds before
-    // calling dispatch — dispatch uses unchecked `std::get` so a
-    // mismatch here would throw `std::bad_variant_access` rather
-    // than our nicer type_mismatch_error.
-    for (std::size_t i = 0; i < arg_count; ++i) {
-      bool ok = false;
-      switch (entry.arg_kinds[i]) {
+    auto kind_matches = [](registry::parser_value const &v,
```
petlenz, line 497: **[Medium — coverage gap] There is no `t2s` arg_kind, so t2s-valued arguments match no overload.** `log(trace(A))` fails with "no overload of 'log' matches" even though log of a scalar invariant is perfectly sensible (trace(A) is t2s). Pre-existing, but the new tensor overloads make it much more surprising — a user who just learned `log(A)` works will expect `log(trace(A))` to work too. Consider either a `t2s` arg_kind that scalar-family functions also accept (t2s is scalar-valued), or a t2s→scalar promotion at the call boundary. At minimum, document the limitation.

src/numsim_cas/parser/actions.h
```diff
@@ -496,37 +494,58 @@ template <> struct action<grammar::function_call> {
...
-        ok = std::holds_alternative<registry::index_list_value>(args[i]);
+        return std::holds_alternative<registry::index_list_value>(v);
+      }
+      return false;
+    };
+
+    // Match on arity first, then on arg kinds — the kind match doubles as the
+    // type check (dispatch uses unchecked std::get, so the chosen overload's
+    // kinds are guaranteed to fit).
+    registry::function_entry const *chosen = nullptr;
```
petlenz, line 518: **[Low — robustness] No ambiguity guard.** Resolution picks the first all-kinds match while iterating an unordered_multimap (unspecified order). Safe today because no two overloads of a name share arg_kinds — but if a future overload collides, the winner becomes silently order-dependent. Cheap hardening: assert at most one candidate matches (or reject duplicate-signature registration), so a collision fails loudly instead of nondeterministically.

src/numsim_cas/parser/function_registry.h
```diff
@@ -221,6 +224,43 @@ inline std::size_t to_positive_size_t(scalar_expr const &e,
...
+    throw type_mismatch_error(std::string{fn_name} +
+                                  ": index must be a non-negative integer "
+                                  "literal",
+                              /*pos=*/0, /*source=*/"");
+  }
+  return static_cast<std::size_t>(*val);
+}
+
+// Spectral-accessor entry: (tensor, integer-index) → t2s eigenvalue.
+inline function_entry eigen_accessor_t2s(auto fn) {
```
petlenz, line 259: **[Low — error UX] Validation errors thrown from dispatch aren't wrapped with parse position.** Verified: `eigenvalue(A, 5)` (out of range) and `eigenvalue(A{rank=4...}, 0)` surface raw `invalid_expression_error` (no line:col), and the non-literal-index path throws `type_mismatch_error` with pos=0 — inconsistent with the positioned errors the parser gives elsewhere. Consider catching cas exceptions in the function_call action and re-throwing with the call's position, and threading `pos` into to_index_size_t (the same #282 gap to_positive_size_t already has).

src/numsim_cas/parser/function_registry.h
```diff
@@ -401,6 +445,30 @@ function_registry() {
...
+    // isotropic tensor log. Resolved by the function_call action's arg-kind
+    // matching (multimap registry).
+    m.emplace("log", tensor_unary([](auto t) { return log(t); }));
+    m.emplace("exp", tensor_unary([](auto t) { return exp(t); }));
+    m.emplace("sqrt", tensor_unary([](auto t) { return sqrt(t); }));
+
+    // Spectral accessors: eigenvalue(A, i) → λ_i (t2s), eigenvector(A, i) → n_i
+    // (rank-1 tensor), eigenprojection(A, i) → E_i = n_i⊗n_i (rank-2 tensor).
+    // The index is a 0-based integer literal.
+    m.emplace("eigenvalue",
```
petlenz, line 483: **[Low — round-trip] The new spectral/invariant names are parse-only.** `eigenvalue(A,0)` prints as `eig_0(A)` (not re-parseable), and `first_invariant(A)` prints as `trace(A)` (lowers to the trace node). That's the same category as the macauley/heaviside lowering caveat documented in this header — but the new names aren't listed there. Add them to the round-trip caveat note so the parse≠print behavior is visible.

petlenz (commented):
Critical self-review of the parser additions. The overload resolver is correct and the new functions work (verified: log(x) scalar vs log(A) tensor resolve; invariants/otimesu/otimesl/spectral accessors land in the right domains; arity and kind errors fire). Findings below, ordered by severity — the two Medium ones I'd act on are the test-strength gap and the t2s-argument gap. Empirically checked, not from memory.

petlenz:
All five review findings addressed in 71fbdbc:

- **#1 (test strength)** — tests now pin each function to its NODE via hash: the three invariants asserted pairwise-distinct (and `I1==trace`, `I3==det` locked); `otimesu`/`otimesl`/`otimes` pairwise-distinct; `eigenvalue` index-sensitivity checked. A swapped/mis-wired dispatch is now caught.
- **#2 (t2s args)** — new `arg_kind::tensor_to_scalar` + t2s overloads: `log(trace(A))`, `sqrt(det(A))`, `exp(norm(A))` now parse (t2s).
- **#3 (ambiguity)** — resolver scans all candidates and throws `internal_error` on a duplicate signature.
- **#4 (positioned errors)** — the parser's own literal-index error is now positioned. cas construction errors (`invalid_expression_error`) are deliberately left to propagate with their type/message — that contract is pinned by the `ParserGrammar` levi_civita/permute_indices tests (my first cut over-wrapped them and broke 4 tests; corrected).
- **#5 (round-trip doc)** — the new parse-only names are listed in the registry header caveat.

1611 tests pass.

petlenz (commented):
Re-review after the fix commit (71fbdbc). The five earlier findings are resolved and the tests are meaningfully stronger (hash identity + sibling discrimination, t2s args, positioned literal-index error, construction-error propagation). One RESIDUAL from my own fix worth noting below; otherwise no new blocking findings.

src/numsim_cas/parser/actions.h
```diff
@@ -496,37 +494,93 @@ template <> struct action<grammar::function_call> {
...
+            expression_holder<tensor_to_scalar_expression>>(v);
       case registry::arg_kind::index_list:
-        ok = std::holds_alternative<registry::index_list_value>(args[i]);
-        break;
+        return std::holds_alternative<registry::index_list_value>(v);
+      }
+      return false;
+    };
+
+    // Match on arity first, then on arg kinds — the kind match doubles as the
```
petlenz, line 513: **[Low — residual from the ambiguity fix] The ambiguity guard is RESOLUTION-time, not REGISTRATION-time.** It throws `internal_error` only when a call actually matches two same-signature entries — so a duplicate-signature registration bug for a name/signature that no test parses would ship latent and only fire in production on the first such call. A stronger guard: validate the multimap once at construction (for each name, assert no two entries share arg_kinds), so the bug fails at first `function_registry()` access regardless of inputs. The resolution-time check is a fine belt; this would be the suspenders.

petlenz:
Residual addressed in 213cf00: added a **registration-time** ambiguity guard — the registry build now asserts no name carries two overloads with identical arg_kinds (throws `internal_error`), catching a duplicate-signature bug at first access regardless of what any test parses. Registry loads clean; 1611 tests pass.

PR #333, opened 2026-07-22
Remove stale TODO(#268): pow-of-pow double-negative already folds

### LINKED ISSUE #268: Scalar pow simplifier: pow(pow(x,-a),-b) → pow(x,a*b) for x,a,b>0
## Summary

`scalar_simplifier_pow.h:53` and `scalar_simplifier_pow.cpp:16` both have:

```cpp
/// TODO? pow(pow(x,-a), -b) --> pow(x,-a*b) only when x,a,b>0
```

The `?` marks this as a speculative future rule. Filing for tracking so the decision (implement vs reject) is explicit.

## The rule

For `x > 0`, `a > 0`, `b > 0`: `pow(pow(x, -a), -b) = pow(x, (-a)(-b)) = pow(x, ab)`.

The current simplifier doesn't combine nested powers when both exponents are negative, even if positivity is known.

## Why the positivity gate matters

For general `x` (no positivity assumption), `pow(pow(x, -a), -b)` can have multiple branches in the complex plane, so the combination is unsafe. The fold is only valid when `x > 0` (real positive base, single-valued power).

## Implementation outline (if accepted)

1. In the `pow_pow_dispatch` simplifier (around `simplifier_pow.h`), detect when:
   - LHS exponent is negative (a scalar tagged `negative` or a literal `<0` constant)
   - Outer exponent is also negative
   - The innermost base has `positive{}` assumption (via `assumptions().contains(positive{})`)
2. Replace with `pow(base, mul(negate(inner_exp), outer_exp))`.
3. Lock-in tests: positive base → fold fires; missing positivity → fold doesn't fire (regression guard against accidental over-folding).

## Decision needed

Should this be implemented as part of #260 (T2S operator positivity propagation) infrastructure, or kept separate? The rule depends on the `positive{}` assumption flowing through, which #260 is meant to enable.

## Priority

Low — speculative. The `?` in the original TODO suggests the original author was unsure if it was worth it. Concrete user case would help.

## Reference

Surfaced during a 1.0-α housekeeping sweep for un-tracked TODOs.

petlenz:
Comment-only cleanup surfaced by the simplifier-issue review.

The `TODO(#268)` in `scalar_simplifier_pow` proposed folding `pow(pow(x,-a),-b) → pow(x,a*b)` *behind a positivity gate*. But the general `pow(pow(x,a),b) → pow(x, a*b)` rule already multiplies exponents **unconditionally** (`scalar_simplifier_pow.cpp:19,23`), so `(-a)(-b)=ab` already fires — and it's test-locked on an unassumed base at `ScalarExpressionTest.h:161-163` (incl. `pow(pow(y,-2),-2) → pow(y,4)`). The proposed gate is unnecessary (real-valued design, no branch cuts) and would *regress* those tests.

Replaces the misleading TODO with an accurate note. No behavior change.

Closes #268.

### DIFF
--- a/include/numsim_cas/scalar/simplifier/scalar_simplifier_pow.h
+++ b/include/numsim_cas/scalar/simplifier/scalar_simplifier_pow.h
@@ -50,7 +50,8 @@ public:
   pow_pow(expr_holder_t lhs, expr_holder_t rhs);
 
   /// pow(pow(x,a),b) --> pow(x,a*b)
-  /// TODO(#268): pow(pow(x,-a), -b) --> pow(x,a*b) only when x,a,b>0
+  /// (The double-negative case pow(pow(x,-a),-b) = pow(x,ab) is covered by
+  /// the a*b exponent multiply above; closed #268.)
   template <typename Expr> expr_holder_t dispatch(Expr const &);
 
   expr_holder_t dispatch(scalar_negative const &rhs);
--- a/src/numsim_cas/scalar/simplifier/scalar_simplifier_pow.cpp
+++ b/src/numsim_cas/scalar/simplifier/scalar_simplifier_pow.cpp
@@ -13,7 +13,8 @@ pow_pow::pow_pow(expr_holder_t lhs, expr_holder_t rhs)
       m_lhs_node{base::m_lhs.template get<scalar_pow>()} {}
 
 /// pow(pow(x,a),b) --> pow(x,a*b)
-/// TODO(#268): pow(pow(x,-a), -b) --> pow(x,a*b) only when x,a,b>0
+/// (The double-negative case pow(pow(x,-a),-b) = pow(x,ab) is covered by
+/// the a*b exponent multiply above; closed #268.)
 template <typename Expr>
 pow_pow::expr_holder_t pow_pow::dispatch(Expr const &) {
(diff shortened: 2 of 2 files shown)


petlenz:
LGTM — comment-only, and the reasoning checks out end to end.

Verified:
- `pow_pow::dispatch` multiplies exponents unconditionally (`pow(x, expr_rhs()*m_rhs)`) in both the generic and `scalar_negative` overloads, so `(-a)(-b)=ab` folds with no gate.
- The cited lock is real and on an **unassumed** base: `ScalarExpressionTest.h:161-163` pins `pow(pow(y,-1),-1)→y`, `pow(pow(y,-2),-2)→pow(y,4)`, `pow(pow(y,-2),2)→pow(y,-4)`. A positivity gate would regress line 162, so the TODO was genuinely stale — the note + closing #268 are justified.

FYI only (pre-existing, not this PR): the parent rule `pow(pow(x,a),b)→pow(x,ab)` is unconditional and not strictly real-domain-sound in general (e.g. `pow(pow(x,2),½)` folds to `x` but is `|x|` for `x<0`). #333 neither introduces nor worsens that; the double-negative sub-case it documents stays sign-consistent. No reason to hold.

PR #402, opened 2026-07-24
Fix #349: overflow-safe scalar_number arithmetic

petlenz:
## Summary

Fixes #349. Stacked on #401.

| Before | After |
|---|---|
| `pow(10,30)` → `5076944270305263616` (mod 2⁶⁴, silent) | `1e30` (double demotion) |
| `rat_add` at ~2⁴⁰ magnitudes → signed-overflow **UB** (UBSan abort) | checked; demotes to double, value ≈ correct |
| `1/2 ÷ 0` → stored `1/0` rational (invariant violation) | `+inf` double, matching the int/int path |
| `normalize_rational`/negation at INT64_MIN → UB | demotes to double |

## Design

- Overflow-checked int64 helpers: `__builtin_*_overflow` on GCC/Clang, manual range checks on MSVC (which is in the CI matrix and lacks both the builtins and `__int128`).
- On overflow the operation **demotes to double** — value stays correct, exactness is lost only at the extremes. In-range exact arithmetic is unchanged and lock-in tested (`1/3 + 1/6` stays the exact rational `1/2`).
- `pow` needed no change: it composes `operator*`, so the checked multiply fixes the wraparound for free.
- Related: #142 fixed the comparison side (`rat_less`) with the same demotion philosophy; this completes the arithmetic side. The follow-up idea of keeping exactness via `__int128` where available is left out deliberately (MSVC portability, and no physics workload needs exact >2⁶³ rationals).

## Tests

4 lock-ins: pow wraparound, rational-add UB repro (value ≈ 2.0), div-by-zero → +inf double, exactness preservation. Full suite: **2437/2437 pass**. gcc-14 `-Werror` compile of the TU clean. The UBSan CI leg (once #356's `-fno-sanitize-recover` lands) becomes the standing regression net for this class.

Issue #356, opened 2026-07-24
CI cannot fail on UBSan findings (recover-by-default) or clang-tidy warnings (WarningsAsErrors empty)


petlenz:
## Summary

Two CI jobs are configured so they can never fail on the class of defect they exist to catch:

1. **UBSan half of the sanitizer leg is non-fatal.** `CMakeLists.txt:195-200` (and the parser copy at 275-280) enable `-fsanitize=address,undefined` but never pass `-fno-sanitize-recover`, and the workflow's Test step sets no `UBSAN_OPTIONS`. GCC UBSan recovers by default: it prints a runtime error and exits 0, so ctest stays green on real UB. (Real UB exists today — see the scalar_number overflow issue #349 — so the moment this is fixed the job should catch regressions.)
2. **Clang-Tidy job cannot fail on warnings.** `.clang-tidy` sets `WarningsAsErrors: ''` and `.github/workflows/clang-tidy-check.yml` runs plain `clang-tidy-18 -p build`; clang-tidy exits 0 when it emits only warnings, so the job only fails on hard compile errors — never on any `bugprone-*`/`performance-*` finding.

Severity: high (silent CI, one-line fixes). Confirmed by config inspection.

## Fix

- Add `-fno-sanitize-recover=undefined` to both sanitizer compile/link blocks (or export `UBSAN_OPTIONS=halt_on_error=1:print_stacktrace=1` in the workflow Test step).
- Add `--warnings-as-errors='*'` to the clang-tidy invocation (or `WarningsAsErrors: '*'` in `.clang-tidy`), triaging the current baseline first so the job starts green.

## Tests / verification

- Temporarily introduce a trivial UB snippet (e.g. signed overflow in a test) on a branch: the sanitizer leg must go red; revert.
- Temporarily introduce a `bugprone-*` violation: the clang-tidy job must go red; revert.
- After #349's fixes land, the sanitizer leg passing becomes the regression test for scalar_number UB.

PR #403, opened 2026-07-24
Fix #361: hash doubles by bit pattern, normalize numerically equal constants

### LINKED ISSUE #361: hash_combine(double) casts to size_t — UB for negative values, all fractional constants collide
## Summary

The generic `hash_combine` overload (`core/hash_functions.h:12-16`) does `static_cast<std::size_t>(value)` on arithmetic types. For `double`:

- **UB for any negative value** — UBSan `-fsanitize=float-cast-overflow` flags `static_cast<size_t>(-2.5)` ("outside the range of representable values"); every negative-double `scalar_constant` is hashed on first comparison/insertion.
- **Gross truncation**: all fractions collide — `hash_combine(seed, 0.5) == hash_combine(seed, 0.9)` (probed), every constant in (0,1) collides with 0.0.

Truncation is "only" a hash-quality problem, but the codebase contains raw hash-equality shortcuts (see #340) and hash-keyed caches, so collisions are not free. Severity: medium.

## Fix

Add a `double` overload that hashes the bit pattern:

```cpp
inline void hash_combine(std::size_t &seed, double v) noexcept {
  if (v == 0.0) v = 0.0;               // normalize -0.0
  hash_combine(seed, std::bit_cast<std::uint64_t>(v));
}
```

## Tests (lock-in)

```cpp
TEST(HashCombine, DoubleNoTruncationNoUB) {
  std::size_t a = 0, b = 0, c = 0, d = 0;
  hash_combine(a, 0.5); hash_combine(b, 0.9);
  EXPECT_NE(a, b);
  hash_combine(c, -2.5);                 // must be UB-free under -fsanitize=float-cast-overflow
  hash_combine(d, 0.0); std::size_t e = 0; hash_combine(e, -0.0);
  EXPECT_EQ(d, e);                       // ±0 normalize together
}
```

Run the suite once under UBSan with float-cast-overflow enabled (see #356).

petlenz:
## Summary

Fixes #361. Stacked on #402. Prerequisite for #356 — the UBSan leg cannot be made fatal while this UB fires on every negative-constant hash.

| Before | After |
|---|---|
| `hash_combine(seed, -2.5)` → `static_cast<size_t>` **UB** (float-cast-overflow) | bit-pattern hash, UB-free |
| `hash(0.5) == hash(0.9)` (all fractions collide with 0) | distinct |
| — | ±0.0 normalize together |

## Design

The subtle part: bit-pattern hashing alone breaks constant folding across variant alternatives — `int64 2` and `double 2.0` are numerically equal (`operator==` promotes) but would hash differently, so `pow(2*sqrt(x),-1) − pow(2.0*sqrt(x),-1)` stopped canceling (caught by `ScalarDifferentiationAudit`). A value-normalizing `hash_combine(scalar_number)` overload fixes this: whole doubles in int64 range hash as their integer; `scalar_constant::update_hash_value` routes through it. Rationals with den==1 already normalize to int64 at construction.

## Tests

2 lock-ins: bit-pattern properties (no truncation, negative-safe, ±0) and cross-alternative equality (hash + fold). Full suite: **2439/2439 pass**. gcc-14 clean.

### DIFF
--- a/include/numsim_cas/core/hash_functions.h
+++ b/include/numsim_cas/core/hash_functions.h
@@ -1,7 +1,9 @@
 #ifndef HASH_FUNCTIONS_H
 #define HASH_FUNCTIONS_H
 
+#include <bit>
 #include <complex>
+#include <cstdint>
 #include <string>
 #include <vector>
 
@@ -13,6 +15,18 @@ inline void hash_combine(std::size_t &seed, const T &value) {
   // std::hash<T> hasher;
   seed ^= static_cast<std::size_t>(value) +
           static_cast<std::size_t>(0x9e3779b9) + (seed << 6) + (seed >> 2);
+}
+
+// Doubles hash by bit pattern: the generic static_cast is UB for negative
+// values and collapses all fractions in (0,1) onto 0 (#361).
+inline void hash_combine(std::size_t &seed, double value) {
+  if (value == 0.0)
+    value = 0.0; // normalize -0.0
+  hash_combine(seed, std::bit_cast<std::uint64_t>(value));
+}
+
+inline void hash_combine(std::size_t &seed, float value) {
+  hash_combine(seed, static_cast<double>(value));
 }
 
 inline void hash_combine(std::size_t &seed, const std::string &value) {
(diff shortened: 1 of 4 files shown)


petlenz:
Landed in main via #409 (merge a5f5ce8), which contained this branch at head dddc4d2. Closing without a separate merge.

PR #404, opened 2026-07-24
Fix #351: rank-4 identity_tensor is major-symmetric, not MinorMajor

petlenz:
## Summary

Fixes #351. Stacked on #403.

`I4_{ijkl} = δ_ik·δ_jl` has **major** symmetry only — swapping i↔j gives δ_jk·δ_il, a different tensor (the minor-symmetric identity is P_sym). The MinorMajor tag propagated (negation, scalar-mul) and routed `inv` through the symmetric 6×6 Voigt path:

| Before | After |
|---|---|
| `inv(-I₄)` at (0,1,0,1) → **−0.25** | −1 (the inverse of −I₄ is −I₄) |
| `is_minor_major(I₄)` → true | false; `is_major(I₄)` → true |

Likely related to the pre-existing rank-4 MinorMajor eval failures tracked in #283/#299.

## Tests

Three tests pinned the wrong annotation (`Rank4MinorIdentityIsMinorMajor` + the move/copy preservation pair from #258's review); updated to Major with the #351 rationale. New numeric lock-in evaluates `inv(-I₄)` end-to-end. `Rank4MinorIdentityIsSymmetric` (is_symmetric via the rank-4 identity) still passes unchanged. Full suite: **2440/2440 pass**. gcc-14 clean.

Issue #382, opened 2026-07-24
[epic] Performance visibility: benchmark harness + evaluation cost model


petlenz:
# [epic] Performance visibility: benchmark harness + evaluation cost model

## Problem

`benchmarks/` exists but contains no meaningful benchmarks; #110 (evaluator microbenchmark), #94 (t2s_eval rebuilt per visit), and #112 (execute the optimization post-bench) have been open since the early phases. Nobody can currently answer "what does evaluating a realistic constitutive tangent cost?", so evaluator design decisions (virtual dispatch per node, `std::any` value transport, per-node spectral decompositions — #375 item 2, n_ary `std::map` allocations, `shared_ptr` refcount traffic) are all unquantified. For a physics-simulation consumer this is the difference between "usable inside an element loop" and "preprocessing only".

## Goal

A benchmark suite in CI (tracked, not gating) that measures the workloads downstream physics actually runs, plus the first round of measured — not guessed — optimizations.

## Scope / tasks

- [ ] Benchmark harness (google-benchmark via FetchContent, `NUMSIM_CAS_BUILD_BENCHMARKS`, default OFF; wire the existing empty `benchmarks/` dir).
- [ ] Representative workloads, committed as code:
  - build + simplify a linear-elasticity energy and its stress/tangent (`examples/linear_elasticity.cpp` as template),
  - `diff(t2s, tensor)` → rank-4 tangent construction time,
  - repeated evaluation of value/stress/tangent at 10⁴–10⁶ points (the element-loop shape),
  - isotropic-function tangent with the divided-difference path (stresses the spectral cache, #375),
  - expression construction/simplification throughput (n_ary merge heavy).
- [ ] Baseline report checked into `docs/` (numbers + machine spec) so regressions are visible in review.
- [ ] First measured optimizations, each justified by a benchmark delta: #94/#112 (t2s_eval rebuild), spectral-cache routing (#375), then whatever the profile actually shows (hash recomputation, map allocations).
- [ ] CI job runs the suite and publishes results as an artifact (no gating initially; add regression thresholds once variance is known).

## Acceptance criteria

- `cmake --build . --target numsim_cas_benchmarks && ./benchmarks/...` produces the workload table locally and in CI.
- #110, #94, #112 closed through this epic with before/after numbers in their PRs.
- Every future "optimization" PR links a benchmark delta.

## Related

#110, #94, #112 (become sub-items), #375, #270 (dim ceiling interacts with workload templates).

[...]

Issue #339, opened 2026-07-24
n_ary_tree operator== ignores the coefficient — x+2 == x+5 and 2*x == 3*x compare equal, corrupting simplify/compare/substitute


petlenz:
## Summary

`n_ary_tree::operator==` (`include/numsim_cas/core/n_ary_tree.h:220-236`) compares hash, id, size, and children — but never `m_coeff`. The n-ary *hash* excludes the coefficient by design (like-term map keying), so with the coefficient also missing from `operator==`, structurally different expressions compare **equal**:

```cpp
auto [x, y] = make_scalar_variable("x", "y");
*(x + 2.0) == *(x + 5.0)      // true
*(2.0*x)   == *(3.0*x)        // true
*sin(x + 2.0) == *sin(x + 5.0) // true (unary_op== recurses into the broken add==)
```

Because many simplifier guards and identity checks use this equality, it produces silently wrong algebra (all reproduced against `build/libNumSim_CAS.a` at 0f6c0b9):

| Input | Output | Correct | Site |
|---|---|---|---|
| `(x+2)*(x+5)` | `pow(2+x,2)` (=9 at x=1) | 18 | `core/simplifier/simplifier_mul.h:28` |
| `sin(x+2)+sin(x+5)` | `2*sin(5+x)` | — | `core/simplifier/simplifier_add.h:44-51` |
| `(x+2) + (-(x+5))` | `0` | `-3` | `scalar/scalar_operators.h:39-47`, `simplifier_add.h:85-90` |
| `eq(2*x, 3*x)` | `1` | indeterminate | `scalar/scalar_std.h:338` |
| `lt(x+2, x+5)` | `0` | identically true | `scalar/scalar_std.h:338` |
| `if_then_else(c, 2*x, 3*x)` | `2*x` | — | `scalar/scalar_std.h:645` |
| `substitute(x+5, x+2, y)` | `y` | no match | `scalar/visitors/scalar_substitution.h:24` |

Same defect in `operator<` (`n_ary_tree.h:198-213`) and in `n_ary_vector.h:151-165`. Related (inverse manifestation, closed): #184. Severity: critical — silent wrong results for everyday inputs; also breaks any user-level `==` / `std::set` over expressions.

## Fix

[...]

PR #405, opened 2026-07-24
Fix #352: substitution no longer inherits the source's space annotation

### LINKED ISSUE #352: substitute() keeps the source expression space annotation — results wrongly tagged Skew/Symmetric, sym/skew/trans then simplify incorrectly
## Summary

`tensor_rebuild_visitor::apply` (`tensor/visitors/tensor_rebuild_visitor.h:29-32`) restores the **source** expression's `space()` onto any rebuilt result lacking one. Sound for pure rebuilds — but `tensor_substitution` inherits it and *changes children*, so the result carries a stale annotation (reproduced at 0f6c0b9):

```cpp
// A, C general rank-2
auto f = substitute(trans(A) - A, trans(A), C);   // result C - A ... tagged Skew!
to_string(skew(f));  // "C - A" unchanged  — WRONG for general C
to_string(sym(f));   // "0{2}"             — WRONG

auto g = substitute(A_sym + B_sym, A_sym, C);     // B + C ... tagged Symmetric
to_string(trans(g)); // returns g unchanged — WRONG
```

Every downstream consumer trusts the tag: projector guards, `trans`, the `inv` skew-singularity check, the symmetry-aware diff kernels. This is a "wrong symmetric flag" soundness hole of the kind #111's preservation matrix wants to test, and adjacent to the closed #93/#297 annotation work.

Severity: high. Confirmed by probe.

## Fix

Make the space restoration conditional: only restore `expr.space()` when the rebuilt node is structurally identical to the source (`m_result == expr`), or add a `protected bool m_preserve_source_space = true;` in the rebuild visitor that `tensor_substitution` sets to false. Substituted results should get their space recomputed from their own structure (structural_propagation), not inherited.

## Tests (lock-in)

```cpp
TEST(Substitution, DoesNotInheritSourceSpace) {
  auto [A, B, C] = make_tensor_variable(std::tuple{"A",3,2}, std::tuple{"B",3,2}, std::tuple{"C",3,2});
  auto f = substitute(trans(A) - A, trans(A), C);   // C - A, general
  EXPECT_NE(to_string(sym(f)), "0{2}");
  EXPECT_NE(to_string(skew(f)), to_string(f));
  // numeric cross-check: eval sym(f) == 0.5*(M + Mᵀ) for M = C_val - A_val
}

TEST(Substitution, RecomputesWhenStructureStillQualifies) {
  auto [A, C] = make_tensor_variable(std::tuple{"A",3,2}, std::tuple{"C",3,2});
  auto h = substitute(trans(A) - A, A, C);          // trans(C) - C — genuinely skew
  EXPECT_EQ(to_string(sym(h)), "0{2}");             // structural skew still detected
}
```

petlenz:
## Summary

Fixes #352. Stacked on #404.

| Before | After |
|---|---|
| `substitute(trans(A)-A, trans(A), C)` → `C-A` tagged **Skew**; `sym` → `0{2}`, `skew` → unchanged | tag dropped; `sym`/`skew` behave correctly |
| `substitute(skew(A), A, B)` keeps Skew (#93 reproducer) | unchanged ✔ |
| `substitute(trans(A)-A, A, B)` → `trans(B)-B` | still skew via construction-time structural detection ✔ |

## Design

The #93 restore in `tensor_rebuild_visitor::apply` is now gated on **structural equality** (`m_result == expr` — honest since #339), with one deliberate exception: **projector contractions** restore across argument substitution, because `skew(X)`/`sym(X)`'s space is derived from the projector, not the argument. That exception is exactly what keeps the original #93 deterministic reproducer green — a plain equality gate broke it, which is how the exception was found.

Note: the equality check runs per rebuilt subtree (recursive `apply`), so worst-case rebuild cost gains an `==` per node; the hash fast-reject makes the changed-subtree case cheap. Flagged for the perf epic's benchmark once #382 lands.

## Tests

2 lock-ins: stale-tag drop (the #352 repro) + operator-derived survival and structural re-derivation. Existing `SkewSpacePreservedThroughRebuild` (#93) passes unchanged. Full suite: **2442/2442 pass**. gcc-14 clean.

### DIFF
--- a/include/numsim_cas/tensor/visitors/tensor_evaluator.h
+++ b/include/numsim_cas/tensor/visitors/tensor_evaluator.h
@@ -385,6 +385,14 @@ private:
   template <typename Op>
   void eval_projector_unary(inner_product_wrapper const &visitable) {
     auto rhs_data = apply(visitable.expr_rhs());
+    // round-3 review: the wrapper's dim() is the projector's; a
+    // dim-changing substitution produced an operand whose buffer this
+    // shortcut then over-read (ASan heap-buffer-overflow)
+    if (rhs_data->dim() != visitable.dim() ||
+        rhs_data->rank() != visitable.rank()) {
+      throw evaluation_error(
+          "projector contraction: operand dim/rank mismatch");
+    }
     m_result = make_tensor_data<ValueType>(visitable.dim(), visitable.rank());
     tensor_data_unary_wrapper<Op, ValueType> op(*m_result, *rhs_data);
     op.evaluate(visitable.dim(), visitable.rank());
--- a/include/numsim_cas/tensor/visitors/tensor_rebuild_visitor.h
+++ b/include/numsim_cas/tensor/visitors/tensor_rebuild_visitor.h
@@ -50,6 +50,9 @@ public:
   // argument, so it survives child substitution (#93/#352).
   static bool same_projector_contraction(tensor_holder_t const &a,
                                          tensor_holder_t const &b) {
+    if (a.get().rank() != b.get().rank() || a.get().dim() != b.get().dim()) {
+      return false; // review on #352: never restore across a shape change
+    }
     auto ia = as_projector_contraction(a);
     auto ib = as_projector_contraction(b);
     return ia && ib && *ia->proj == *ib->proj;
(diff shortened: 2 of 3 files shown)


petlenz:
Landed in main via #409 (merge a5f5ce8), which contained this branch at head a703520. Closing without a separate merge.

PR #406, opened 2026-07-24
Fix #354: t2s constant_mul keeps a symbolic scalar_wrapper factor

### LINKED ISSUE #354: t2s constant_mul drops a non-numeric scalar_wrapper factor: wrapper(x)*(trace(A)*det(A)) → 1*det(A)*tr(A)
## Summary

`constant_mul::dispatch(tensor_to_scalar_mul)` (`src/numsim_cas/tensor_to_scalar/simplifier/tensor_to_scalar_simplifier_mul.cpp:31-44`) silently **drops** a non-numeric `scalar_wrapper` LHS when multiplying into an existing product (reproduced at 0f6c0b9):

```cpp
// x scalar symbol, A tensor
wrapper(x) * (trace(A) * det(A))   // → "1*det(A)*tr(A)" — x has vanished, coeff reset to 1
```

When `lhs_val` (numeric extraction) is `nullopt`, the code computes `coeff{get_coefficient(mul,1)}`, skips the `coeff * *lhs_val` merge, and never inserts `m_lhs` into the product. The promoted path `x * (f*g)` happens to route through `mul_base::dispatch` and is correct — which is why tests miss it; the bug fires whenever the wrapper is the visitor's LHS (e.g. `(x*f)*g` built in an order that puts the wrapper first, or internal simplifier calls).

Severity: high (silent factor loss). Confirmed. Related: #104 (mul×mul merge consolidation — this is a concrete drift instance of that risk).

## Fix

In the `!lhs_val` branch, insert `m_lhs` as a child (`push_or_combine(mul, m_lhs)`, merging with an existing scalar_wrapper child the way `n_ary_add::dispatch(scalar_wrapper)` does) instead of touching only the numeric coefficient.

## Tests (lock-in)

```cpp
TEST(T2sConstantMul, SymbolicWrapperFactorSurvives) {
  auto [x] = make_scalar_variable("x");
  auto [A] = make_tensor_variable(std::tuple{"A", 3, 2});
  auto f = trace(A) * det(A);
  auto e = /* wrapper-first product, e.g. via promote/explicit wrapper */ wrap(x) * f;
  // evaluate with x=2, A=diag(1,2,3): trace=6, det=6 → expect 72, not 36
  EXPECT_NE(to_string(e).find("x"), std::string::npos);
}
```

(Construct the wrapper-first path the same way the probe did — building `tensor_to_scalar_scalar_wrapper(x)` explicitly and multiplying it by an existing `tensor_to_scalar_mul` — so the regression can't dodge through the promoted route.)

petlenz:
## Summary

Fixes #354. Stacked on #405.

`wrapper(x) * (trace(A)*det(A))` → `1*det(A)*tr(A)` — the symbolic factor **x vanished** and the coefficient was reset. When `try_numeric` fails on the LHS wrapper, `constant_mul::dispatch(tensor_to_scalar_mul)` never inserted it. The promoted route (`x * (f*g)`) goes through `mul_base` and was correct, which is why the suite missed it.

## Fix

Non-numeric wrappers insert as factors via `push_or_combine` (forward-declared; merging with an existing wrapper child through the existing unwrap-multiply-rewrap path). Numeric path unchanged.

## Tests

Typed lock-in across all three dims (wrapper-first construction so the regression can't dodge through the promoted route). Full suite: **2445/2445 pass**.

### DIFF
--- a/src/numsim_cas/tensor_to_scalar/simplifier/tensor_to_scalar_simplifier_mul.cpp
+++ b/src/numsim_cas/tensor_to_scalar/simplifier/tensor_to_scalar_simplifier_mul.cpp
@@ -12,6 +12,9 @@ namespace simplifier {
 // --- constant_mul ---
 using t2s_traits = domain_traits<tensor_to_scalar_expression>;
 
+static void push_or_combine(tensor_to_scalar_mul &mul,
+                            mul_base::expr_holder_t const &child);
+
 constant_mul::constant_mul(expr_holder_t lhs, expr_holder_t rhs)
     : base(std::move(lhs), std::move(rhs)),
       lhs_val{t2s_traits::try_numeric(base::m_lhs)} {}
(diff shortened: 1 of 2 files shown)


petlenz:
Landed in main via #409 (merge a5f5ce8), which contained this branch at head 7a945d5. Closing without a separate merge.

Issue #383, opened 2026-07-24
[epic] Evaluation generality: honest ceilings, checked at construction, then lifted where physics needs them


petlenz:
# [epic] Evaluation generality: honest ceilings, checked at construction, then lifted where physics needs them

## Problem

The evaluator has compile-time walls that fail late and opaquely:

- **MaxDim = 3** (#270) — dim-4+ expressions construct fine, then die in the evaluator; blocks e.g. plane-stress 4th-dim tricks and any spacetime formulation.
- **MaxRank = 8** — the cause of the 16 skipped fuzz seeds: a depth-3 random derivative reaches rank 10 and throws `make_tensor_data_imp::mismatch` at evaluation time. Legitimate physics reaches rank 6 routinely (d(rank-4 tangent)/d(rank-2) for sensitivity/higher-order methods → rank 6; rank 8 is the current edge).
- **Pattern-limited contraction evaluation** — `tensor_data_dcontract_wrapper` handles the plain rank-2 `{1,2}/{1,2}` case only (#353); general index-sequence contraction, rank-1 dots, and rank-4:4 contractions either mis-evaluate or throw.
- Construction never checks any of this: users learn the ceilings from runtime `evaluation_error`s (or worse, from #360-class silent acceptance).

## Goal

1. Every ceiling is enforced with a clear message **at construction** ("expression exceeds evaluator support: rank 10 > 8").
2. The contraction evaluator becomes general over index sequences within the supported dim/rank envelope.
3. The envelope itself becomes a documented, deliberate choice — lifted where a physics use case exists, not by default (compile time and data-size grow fast with MaxDim/MaxRank).

## Scope / tasks

- [ ] Construction-time capability check: a `max_evaluable_rank()/dim()` query on the data layer + a guard in the factories (or a `validate_evaluable(expr)` the user can call before entering a hot loop).
- [ ] General index-sequence contraction in `tensor_data_*` (closes the evaluator half beyond #353's minimal fix): permute-then-contract or index-loop, ranks 1–4 operands, any legal sequence pair.
- [ ] Decide + document the shipped envelope (proposal: MaxDim 3, MaxRank 8 kept as defaults; both become CMake-configurable template parameters, with a CI leg compiling MaxDim=4 to keep it honest — addresses #270 without paying its compile cost by default).
- [ ] Un-skip the 16 fuzz seeds under a MaxRank-elevated test configuration, or cap fuzz generation at the queried envelope so skips become impossible by construction.
[...]

Issue #343, opened 2026-07-24
tensor_inner_product_to_scalar hash/equality ignore the index sequences — dot(A,{1,2},B,{1,2}) == dot(A,{1,2},B,{2,1})


petlenz:
## Summary

`tensor_inner_product_to_scalar` (`include/numsim_cas/tensor_to_scalar/tensor_inner_product_to_scalar.h`) stores contraction index sequences but has no `update_hash_value`/`operator==` override — it inherits `binary_op`'s hash (id+lhs+rhs only) and `binary_op::operator==` (`core/binary_op.h:162-166`, children only). Distinct contractions compare equal (reproduced at 0f6c0b9, non-symmetric A, B):

```cpp
dot_product(A, {1,2}, B, {1,2}) == dot_product(A, {1,2}, B, {2,1})  // true — A:B vs A:Bᵀ
// their sum  → 2*dot(A,{1,2},B,{2,1})   — WRONG
// their diff → 0                        — WRONG
```

This is the same bug class as #266 (fixed there for `inner_product_wrapper`; `tensor_to_scalar_eigenvalue.h:42-48` explicitly mirrors that fix — this node was missed) and the permute_indices_wrapper issue (#342).

Severity: critical (silent wrong algebra for any permuted contraction). The evaluator half of this node's problems is tracked separately (evaluator ignores the sequences).

## Fix

Add an `update_hash_value()` override that `hash_combine`s `m_lhs_indices`/`m_rhs_indices` (copy `inner_product_wrapper.h:59-67`), and a member `operator==` comparing both sequences in addition to the children.

## Tests (lock-in)

```cpp
TEST(T2sInnerProductIdentity, IndexSequencesDistinguish) {
  auto [A] = make_tensor_variable(std::tuple{"A", 3, 2});
  auto [B] = make_tensor_variable(std::tuple{"B", 3, 2});
  auto ab  = dot_product(A, sequence{1,2}, B, sequence{1,2});
  auto abt = dot_product(A, sequence{1,2}, B, sequence{2,1});
  EXPECT_FALSE(*ab == *abt);
  EXPECT_NE(ab.get().hash_value(), abt.get().hash_value());
  EXPECT_NE(to_string(ab - abt), "0");
  // identical sequences still fold
  auto ab2 = dot_product(A, sequence{1,2}, B, sequence{1,2});
  EXPECT_EQ(to_string(ab - ab2), "0");
}
```

PR #407, opened 2026-07-24
Fix #353: t2s contraction evaluation honors the index sequences

### LINKED ISSUE #353: t2s evaluator ignores tensor_inner_product_to_scalar index sequences — dot(A,{1,2},B,{2,1}) evaluates as A:B; rank-1/rank-4 contractions throw
## Summary

The t2s evaluator ignores the index sequences stored on `tensor_inner_product_to_scalar` (`tensor_to_scalar/visitors/tensor_to_scalar_evaluator.h:158-165`): it always computes plain `tmech::dcontract(l, r)` via `tensor_data_dcontract_wrapper` (`tensor/data/tensor_data_to_scalar_wrapper.h:56-89`), which is hardcoded to `l_ij·r_ij`, rank 2 only. Reproduced at 0f6c0b9:

- `dot_product(A, {1,2}, B, {2,1})` (= A_ij·B_ji) evaluates to **542.5** (the plain A:B) instead of **479.5**.
- `dot_product(u, {1}, v, {1})` on rank-1 vectors throws `"tensor_data_dcontract_wrapper: requires rank 2"` although the node is a legal full contraction. This also breaks evaluating derivatives of `dot()` for rank-1/rank-4 arguments — `tensor_to_scalar_differentiation_wrt_scalar.cpp:225-244` builds exactly such nodes.

Sibling identity bug (hash/== ignore the same sequences) is #343; this issue is the evaluator half.

Severity: high (wrong values / unevaluable derivatives). Confirmed.

## Fix

Pass `v.indices_lhs()` / `v.indices_rhs()` into the wrapper and perform the general contraction (permute one operand per the sequences before `dcontract`, or index-loop), supporting ranks 1–4. Minimal stopgap: special-case the transposed rank-2 pattern ({1,2}/{2,1} → `dcontract(l, trans(r))`) and the rank-1 dot, and throw a clear `not_implemented_error` otherwise.

## Tests (lock-in)

```cpp
TEST(T2sInnerProductEval, TransposedContraction) {
  // A, B non-symmetric rank-2 with known entries
  // expect eval(dot_product(A,{1,2},B,{2,1})) == sum_ij A(i,j)*B(j,i)
}

TEST(T2sInnerProductEval, Rank1Dot) {
  // u, v rank-1: expect eval(dot_product(u,{1},v,{1})) == u·v (no throw)
}

TEST(T2sInnerProductEval, DiffOfVectorDotEvaluates) {
  // s scalar; u(s) rank-1: diff(dot(u), s) must evaluate (currently throws "requires rank 2")
}
```

petlenz:
## Summary

Fixes #353. Stacked on #406. Completes the node whose identity half landed in #399 (#343).

| Before | After |
|---|---|
| `dot_product(A,{1,2},B,{2,1})` evaluated as plain A : B (542.5 vs correct 479.5 in the issue repro) | contracts against the transpose |
| `dot_product(u,{1},v,{1})` threw `requires rank 2` | `tmech::dot` — also unblocks evaluating `diff(dot(u), s)` for rank-1 args |
| rank>2 silently wrong / opaque throw | clear not-implemented error referencing the #383 ceilings epic (general sequence-driven contraction is that epic's task) |

## Tests

Numeric lock-in comparing both rank-2 orientations against hand-computed sums plus the rank-1 dot. Full suite: **2446/2446 pass**. gcc-14 clean.

### DIFF
--- a/include/numsim_cas/tensor/data/tensor_data_to_scalar_wrapper.h
+++ b/include/numsim_cas/tensor/data/tensor_data_to_scalar_wrapper.h
@@ -4,6 +4,7 @@
 #include "spectral_decomposition_cache.h"
 #include "tensor_data.h"
 #include <numsim_cas/core/cas_error.h>
+#include <numsim_cas/tensor/sequence.h>
 
 #include <algorithm>
 #include <array>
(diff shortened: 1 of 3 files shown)


petlenz:
Landed in main via #409 (merge a5f5ce8), which contained this branch at head 3eb5bf3. Closing without a separate merge.

PR #408, opened 2026-07-24
Fix #350: tensor_pow contract — rank-2 only, integer exponents, working negative powers

### LINKED ISSUE #350: tensor_pow broken beyond positive-int/rank-2: eval(pow(X,-1)) returns X (no inversion), pow(X,0.5) evaluates to identity, diff(pow(X,-n)) silently 0, pow(rank-4, 0) yields rank-2 identity
## Summary

`tensor_pow`'s contract is broken end-to-end for anything but positive integer exponents on rank-2 operands — each layer fails differently and silently (all reproduced at 0f6c0b9):

1. **Evaluator, negative exponent → no inversion.** `operator()(tensor_pow)` (`tensor/visitors/tensor_evaluator.h:257-288`) does `static_cast<int>(exp_val)` then contracts `k < std::abs(n)` times — `pow(X,-1)` evaluates to **X itself** (probe: got 2.0 where inverse gives 0.5).
2. **Evaluator, fractional exponent → identity.** `pow(X, 0.5)` truncates to n=0 and silently returns the identity.
3. **diff, negative exponent → silent zero.** The product-rule loop `for r = 0; r < n` never runs for n<0; the invalid sum coerces to zero: `diff(pow(X,-2), X)` → `0{4}`. Sites: `src/numsim_cas/tensor/visitors/tensor_differentiation.cpp:44-72`, `tensor_differentiation_wrt_scalar.cpp:56-75`. Corrupts any tangent containing a negative matrix power.
4. **Factory, rank/exponent gates missing.** `pow(C4, 0)` yields a **rank-2** identity for a rank-4 operand (`tensor/tensor_std.h:46-52` hard-codes `identity_tensor(dim, 2)`), and `pow` accepts any rank although the evaluator's repeated single-index contraction and the diff kernel are only meaningful for rank-2.

Severity: high. Related: #287 (non-constant exponent diff — separate deferred case), #192 (the same missing-gate pattern for `inv`), #96.

## Fix

Smallest coherent change at the factory (`tensor_std.h::pow`):
- gate to rank-2 operands (`invalid_expression_error`, matching `dev`/`sym`/`trans` and the #192 fix for `inv`);
- rewrite `pow(A, -n)` → `inv(pow(A, n))` for negative integer constants (evaluator and diff already handle `inv` correctly);
- reject non-integer constant exponents (or route to the isotropic-function machinery of #227).
Defense in depth: evaluator throws `evaluation_error` on non-integer `exp_val`; both diff visitors throw `not_implemented_error` for n<0 instead of returning an invalid (→ zero) sum.

## Tests (lock-in)

```cpp
TEST(TensorPow, NegativeExponentInverts) {
  auto [X] = make_tensor_variable(std::tuple{"X", 3, 2});
  tensor_evaluator<double> ev;  // X = diag(2,4,5)
  auto r = ev.apply(pow(X, -1));
  EXPECT_NEAR(r->raw_data()[0], 0.5, 1e-12);   // was 2.0
}

TEST(TensorPow, FractionalExponentRejected) {
  auto [X] = make_tensor_variable(std::tuple{"X", 3, 2});
  EXPECT_THROW(pow(X, 0.5), cas_error);         // or evaluation_error at eval — never silent identity
}

TEST(TensorPow, DiffNegativeExponentNotZero) {
  auto [X] = make_tensor_variable(std::tuple{"X", 3, 2});
  auto d = diff(pow(X, -2), X);
  EXPECT_NE(to_string(d), "0{4}");
  // numeric FD cross-check at X = diag(2,4,5) (FuzzyTensorDiffTest harness style)
}

TEST(TensorPow, RankGate) {
  auto [C] = make_tensor_variable(std::tuple{"C", 3, 4});
  EXPECT_THROW(pow(C, 2), invalid_expression_error);
  // and pow(X, 0) for rank-2 returns rank-2 identity (unchanged)
}
```

petlenz:
## Summary

Fixes #350. Stacked on #407.

| Before | After |
|---|---|
| `eval(pow(X,-1))` returned **X itself** (no inversion) | inverts: 0.5 for diag entry 2 |
| `pow(X, 0.5)` evaluated to the **identity** (int truncation) | factory rejects; evaluator throws on non-integer values |
| `diff(pow(X,-2), X)` → silent `0{4}` | `not_implemented_error` pointing at `inv(pow(A,n))` (which differentiates correctly today) |
| `pow(C4, 0)` → **rank-2** identity for a rank-4 operand | rank gate at the factory (`invalid_expression_error`) |

## Design

Construct-early philosophy per the #383 ceilings epic: the factory gates rank (2 only — matching `dev`/`sym`/`trans` and #192's `inv` fix) and integer-constant exponents; the evaluator and diff visitors keep defense-in-depth guards for trees built by other paths. Negative integer powers now evaluate correctly (invert the accumulated power); their differentiation stays explicit-unsupported rather than silently wrong — the `inv(pow(A,n))` spelling covers that need (lock-in included).

## Tests

3 lock-ins: gates, numeric inverse evaluation (n=−1, −2), diff throw + supported spelling. Full suite: **2449/2449 pass**. gcc-14 clean.

### DIFF
--- a/include/numsim_cas/tensor/tensor_std.h
+++ b/include/numsim_cas/tensor/tensor_std.h
@@ -48,7 +48,10 @@ template <tensor_expr_holder ExprLHS, scalar_expr_holder ExprRHS>
   }
   // Non-integer constant exponents have no matrix-power meaning here;
   // isotropic tensor functions (#227) cover fractional powers (#350).
-  if (is_same<scalar_constant>(expr_rhs)) {
+  if (is_same<scalar_constant>(expr_rhs) ||
+      (is_same<scalar_negative>(expr_rhs) &&
+       is_same<scalar_constant>(
+           expr_rhs.template get<scalar_negative>().expr()))) {
     if (!try_int_constant(expr_rhs)) {
       throw invalid_expression_error(
           "pow: tensor exponent must be an integer constant");
--- a/include/numsim_cas/tensor/tensor_std.h
+++ b/include/numsim_cas/tensor/tensor_std.h
@@ -39,6 +39,21 @@ namespace numsim::cas {
 
 template <tensor_expr_holder ExprLHS, scalar_expr_holder ExprRHS>
 [[nodiscard]] auto pow(ExprLHS &&expr_lhs, ExprRHS &&expr_rhs) {
+  // Matrix powers are rank-2 only; the evaluator's repeated contraction
+  // and the diff kernels are meaningless for other ranks (#350).
+  if (expr_lhs.get().rank() != 2) {
+    throw invalid_expression_error(
+        "pow: tensor operand must be rank 2 (got rank " +
+        std::to_string(expr_lhs.get().rank()) + ")");
+  }
+  // Non-integer constant exponents have no matrix-power meaning here;
+  // isotropic tensor functions (#227) cover fractional powers (#350).
+  if (is_same<scalar_constant>(expr_rhs)) {
+    if (!try_int_constant(expr_rhs)) {
+      throw invalid_expression_error(
+          "pow: tensor exponent must be an integer constant");
+    }
+  }
   // pow(0, n) → 0
   if (is_same<tensor_zero>(expr_lhs))
     return make_expression<tensor_zero>(expr_lhs.get().dim(),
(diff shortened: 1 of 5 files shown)


petlenz:
Landed in main via #409 (merge a5f5ce8), which contained this branch at head 9fa3067. Closing without a separate merge.

